Issue #1, opened 2021-04-22
Hello world rust executable, with a test suite.


soenkehahn:
The test can just be something like `assert_eq!(true, true)`.

PR #7, opened 2021-04-23
add 'hello world' executable

soenkehahn:


soenkehahn:
Fixes #1.

rust-toolchain
```diff
@@ -0,0 +1 @@
+1.51.0
```
soenkehahn, line 1: Is there a better way to pin down the rust version than this? @casey

PR #8, opened 2021-04-23
Serve hello world html page

### LINKED ISSUE #6: Hello world server
Delivers a server-rendered html page that says 'hello world'

soenkehahn:
Fixes #6.


src/main.rs
```diff
@@ -1,11 +1,98 @@
...
+            head {
+                title {
+                    "foo"
+                }
+            }
+            body {
+                h1 { "hello world" }
+            }
+        }
+    };
```
soenkehahn, line 42: I've used `maud` for html rendering now. Not sure whether I like that though.

Cargo.toml
```diff
@@ -5,3 +5,14 @@ authors = ["Casey Rodarmor <casey@rodarmor.com>", "Sönke Hahn <soenkehahn@gmail
 edition = "2018"
 
 [dependencies]
+maud = "0.22.2"
```
casey, line 8: How about an empty line after this line, just for cosmetic purposes?

casey:
Nice, looks good!

The compile-time template library that I used before is [askama](https://github.com/djc/askama). It's more of a traditional text-template engine, with a `templates` directory full of template files, which are mostly text aside from `{{...}}` interpolations.

To use a template, you declare a struct like this:

```rust
#[derive(Template)]
#[template(path = "hello.html")]
struct Foo {
  bar: u32,
  baz: String,
}
```

And can then render it like so:

```rust
Foo { bar: 11, baz: "foo".into() }.render().unwrap()
```

I see maud is a bit more like JSX, with inline macros that generate HTML. I've never used it, but it and askama have a similar amount of stars, and the code looks good.  I kind of prefer the askama approach, and like the separation of template files and code. But, it isn't a strong preference. And since I haven't used maud, I won't mind trying it out and seeing how it feels. It shouldn't be too hard to switch later if we need to.

What do you think?

casey:
(If you want to go with maud, feel free to merge as is.)

Cargo.toml
soenkehahn: Good catch. This was `cargo add`'s doing.

PR #9, opened 2021-04-24
Add rustfmt config

casey:
Add a rustfmt config. I usually use nightly rustfmt, since it supports a bunch of nice options, but that requires editor configuration. (For example, [here's Just's rustfmt.toml](https://github.com/casey/just/blob/master/rustfmt.toml). I think it's worth it, but that's debatable, so I thought I'd start by adding a stable config.


rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+edition = "2018"
+max_width = 100
```
soenkehahn, line 2: This seems to be the default. Should we only overwrite values that we want to not be the default?

rustfmt.toml
```diff
@@ -0,0 +1,6 @@
...
+newline_style = "Unix"
+tab_spaces = 2
```
soenkehahn, line 4: Sounds good to me. :)

soenkehahn (commented):
Apart from removing the options that just reiterate the defaults, LGTM!

PR #11, opened 2021-04-24
Add Github Actions CI

casey:
Add github actions CI config. When we need it, I can also add code to package and upload releases.


.github/workflows/build.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - master
+    tags:
+    - '*'
+  pull_request:
+    branches:
+    - master
+
+defaults:
+  run:
+    shell: bash
```
casey, line 15: This makes the windows build default to bash, instead of, I think, PowerShell.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+on:
+  push:
+    branches:
+    - master
+    tags:
+    - '*'
+  pull_request:
+    branches:
+    - master
```
casey, line 11: Runs CI on master, all tags, and pull requests against master.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  # Increment this key to invalidate the cache.
+  CACHE_KEY: 0
```
casey, line 19: GitHub doesn't provide a way to invalidate caches. Later, this variable is used as part of the cache key, so incrementing it will cause a cache miss. Useful if there's a problem with the cache, which I've run into before.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+    name: All
+
+    strategy:
+      matrix:
+        os:
+        - macos-latest
+        - ubuntu-latest
+        - windows-latest
+
+    runs-on: ${{matrix.os}}
```
casey, line 32: Run the build on Linux, MacOS, and Windows.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    env:
+      RUSTFLAGS: "-D warnings"
```
casey, line 35: Elevate all warnings to errors.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
```
casey, line 38: This step checks out the code.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    # An issue with BSD Tar causes sporadic failures on macOS.
+    # c.f https://github.com/actions/cache/issues/403
+    - name: Install GNU Tar
+      if: matrix.os == 'macos-latest'
+      run: |
+        brew install gnu-tar
+        echo /usr/local/opt/gnu-tar/libexec/gnubin > $GITHUB_PATH
```
casey, line 46: Installs GNU Tar and adds it to the path, to fix some cache nonsense on macos.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    - name: Cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: cargo-${{ env.CACHE_KEY }}-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}
```
casey, line 55: Set up the cache per OS, and invalidate it if any Cargo.lock file changes. Invalidating it on any Cargo.lock file change seems aggressive to me, but it's what the documentation suggests.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    - name: Install Rust Toolchain Components
+      run: |
+        rustup component add clippy rustfmt
```
casey, line 59: Rustup is already present, and the right toolchain will be installed automatically whenever `cargo` runs, because of the `rust-toolchain` file, but clippy and rustfmt need to be installed manually.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+    - name: Info
+      run: |
+        rustup --version
+        cargo --version
+        cargo clippy --version
+        cargo fmt --version
```
casey, line 66: Print out info on toolchain and components. Can be ignored, but useful if you need to know what version of everything is running.

.github/workflows/build.yaml
soenkehahn: Maybe `--deny`?

PR #19, opened 2021-04-26
List directory entries in `./www`

casey:
I found [this example](https://github.com/hyperium/hyper/blob/master/examples/tower_server.rs), which shows how to construct a service without using `make_service_fn` or `service_fn`, and used it to refactor the code into two services, `ConnectionHandler` and `RequestHandler`.

The `ConnectionHandler` is called on each incoming connection, and constructs a `RequestHandler` to handle the requests.

I'm not super attached to either of these names, so let me know if you have better ideas.

I replaced `impl Future` with a typedef:

```
type ConnectionHandlerServer = Server<AddrIncoming, ConnectionHandler>;
```

This allows function signatures to be a little nicer and proves errors. Also, since `Server` exposes the port, it allows us to get rid of `(port, future)` tuples.

I didn't bother doing any proper error handling, but made a card noting that we should do that.

Before merging, I would put `ConnectionHandler` and `RequestHandler` into `src/connection_handler.rs` and `src/request_handler.rs` respectively, and move common includes into `src/common.rs`, but we didn't talk about how to structure the project, so I'm fine leaving that for later.


soenkehahn:
I think proper error handling should definitely not be deferred, if it's needed. We should try to not have anything on master that does things (e.g. error handling) in a non-optimal way. So I think we should introduce error handling (including libraries for that) in this PR as needed. (Maybe right now we should go with something simple like `anyhow` and switch to something more elaborate once we actually have some requirements for error messages. Although I think we might want to include a test for the case where the `www` directory doesn't exist?)

I like the typedef!

And I don't like hyper. :| I think this pattern of having two server implementations, one for handling incoming connections and one for handling incoming requests is pretty ugly. At least when they both use the same trait. Not sure whether we should switch to something better now though.

soenkehahn:
Also, maybe we should use an `async` way to access the file system.

src/main.rs
```diff
@@ -1,72 +1,76 @@
...
-  let server = Server::bind(&addr).serve(make_service);
-  let port = server.local_addr().port();
-  eprintln!("listening on port {}", port);
-  (port, server)
-}
-
-async fn run(server: impl Future<Output = Result<(), impl Display>>) {
-  if let Err(e) = server.await {
-    eprintln!("server error: {}", e);
+  match run().await {
```
casey, line 9: How about `if let Err(error) => {` since we don't do anything with `Ok`?

src/request_handler.rs
```diff
@@ -0,0 +1,124 @@
...
+  fs,
+  io::{self, Cursor, Write},
+  net::SocketAddr,
+  path::{Path, PathBuf},
+  sync::{Arc, Mutex},
+  task::{self, Poll},
+};
+use tower::make::Shared;
+
+#[derive(Clone, Debug)]
```
casey, line 19: How about putting Stderr in its own module?

casey:
A couple minor nits, but otherwise looks good to me!

soenkehahn:
I pushed some more commits. I think they make the PR even nicer, let me know what you think. (Looking at them commit by commit might be the easiest.)

casey:
Nice, looks good

PR #21, opened 2021-04-26
Link with lld for a faster edit-compile-test cycle

soenkehahn:
This changes the time it takes to run the test suite from `~2.2` seconds to `~0.7` on my machine. (On `master`.)


soenkehahn:
Fixes #14.

soenkehahn:
I'll put this in `~/.cargo/config.toml` instead.

PR #22, opened 2021-04-26
Move ci commands into `bin/ci.sh` to be able to run it locally

soenkehahn:
(@casey: We talked about this a few days ago.)

I ran into the situation where I wanted to run all the ci commands locally to fix the ci failures on a branch. It'd be much easier to do that with this locally available script. I'm also creating this pull request so that we can see, how exactly it looks like in the github actions web ui.


.github/workflows/build.yaml
```diff
@@ -65,14 +62,5 @@ jobs:
...
-
-    - name: Test
-      run: cargo test --all
-
-    - name: Clippy
-      run: cargo clippy --all
-
-    - name: Check Formatting
-      run: cargo fmt --all -- --check
+    - name: CI
```
soenkehahn, line 65: Can we come up with a better name than `CI`?

bin/ci.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -eux
```
soenkehahn, line 3: `-x` makes bash log every command before executing it. So that we can see exactly which steps produced which output (or failed) on github.

PR #23, opened 2021-04-27
Add justfile to run ci-like checks locally instead of ci script

soenkehahn:
The problem with the existing ci script is that it set `RUSTFLAGS`. This affects the compilation of all of our dependencies. So `cargo` will recompile all of them. For example (starting with no `target` directory) running `cargo test`, followed by `./bin/ci`, followed by `cargo test` will build all dependencies three times. This makes using the ci script locally very unappealing.

(It's all a bit sad, because we only use `RUSTFLAGS` to set `--deny warnings`, but that setting will somehow be overridden for dependencies by `cargo`. So it's not at all useful to rebuild those dependencies.)

This PR reverts #22 and also adds a `justfile` to serve a similar purpose: It allows to run all checks that are run on ci, with the difference of not denying warnings.

The downside of this approach is that we probably want to keep the checks in `justfile` and in `.github/workflows/build.yaml` in sync manually.


casey:
LGTM

.github/workflows/build.yaml
```diff
@@ -62,5 +65,14 @@ jobs:
...
-    - name: CI
-      run: ./bin/ci
+    - name: Build
+      run: cargo build --all
+
+    - name: Test
+      run: cargo test --all
+
+    - name: Clippy
+      run: cargo clippy --all
```
xasopheno, line 75: I've 
It can be fun to `+nightly` and get all the new lints.
```
clippy:
	cargo +nightly clippy --all-targets
```
I also have this in my `src/lib`:
```
#![warn(
    clippy::nursery,
    //clippy::restriction,
    //clippy::pedantic,
    //clippy::cargo,
)]
```
I only use the nursery lints, but it can be fun to see the other lints so I turn them on and off from time to time. 
All of this said, it's been a while since I've poked around in clippy land. There might be some new, cool ways of doing the same thing.

Issue #25, opened 2021-04-27
Handle files with non-unicode filenames correctly.


soenkehahn:

casey:
I think we should just refuse to serve filenames that aren't valid unicode.

- We can use a crate like [camino](https://docs.rs/camino/1.0.5/camino/) to represent paths. `camino` paths implement `Display` directly and no methods return types like `&OsStr` or `OsString`
- Filenames that aren't valid unicode won't display in a useful way when downloaded onto a user's machine. I'm actually not sure what they would display as
- There isn't much of a reason to support non-unicode paths. Other encodings aren't very popular

soenkehahn:
Sounds good to me.

casey:
I changed the title and hijacked the top comment. I kinda wish the top comment wasn't attributed to anyone, since I think it's nice to change it as needed to reflect what the current understanding of the issue is, or have a checklist for what needs to be done to close the issue.

PR #29, opened 2021-04-28
Allow to download files

soenkehahn:


src/request_handler.rs
```diff
@@ -59,6 +69,16 @@ impl RequestHandler {
 
     Ok(Response::new(Body::from(body.into_string())))
   }
+
+  async fn serve_file(&self, file_path: &str) -> Result<Response<Body>> {
+    // todo: stream files
```
soenkehahn, line 74: Currently we're loading whole files into memory. Which is probably bad.

src/request_handler.rs
```diff
@@ -25,8 +25,8 @@ impl RequestHandler {
     }
   }
 
-  async fn response(mut self) -> Response<Body> {
-    match self.list_www().await.context("cannot access `www`") {
+  async fn response_without_errors(mut self, request: Request<Body>) -> Response<Body> {
```
casey, line 28: I kinda don't like this name, although don't feel super strongly about it.

I think I'd slightly prefer keeping this called `response`, and calling the fallible function that's currently called `response` `dispatch` instead, since it looks like it's responsible for doing routing. What do you think?

src/request_handler.rs
```diff
@@ -38,6 +38,14 @@ impl RequestHandler {
     }
   }
 
+  async fn response(&self, request: Request<Body>) -> Result<Response<Body>> {
+    dbg!(request.uri().path());
```
casey, line 42: This should be removed before merging. (Although perhaps this anticipates adding the check for forbidden words, to make sure it works?!?)

src/request_handler.rs
```diff
@@ -50,7 +58,9 @@ impl RequestHandler {
         }
         body {
           @while let Some(entry) = read_dir.next_entry().await? {
-            (entry.file_name().to_string_lossy())
+            a download href=(format!("./{}", entry.file_name().to_string_lossy())) {
```
casey, line 61: This will have weird behavior for non-unicode file names. They'll appear in listings, but you'll get a 404 when you click them, because the client will try to retrieve them with a path that includes the unicode replacement character (�) instead of the actual bytes in the path.

I think what we should probably do right now is just not list files with non-unicode paths, and somehow warn the user about the situation.

We could somehow generate links that contained a representation of the path (like hex) and decoded it when the file was retrieved. However, I think in general it's probably best to just refuse to handle non-unicode text, and force the user to deal with it, since it avoids a ton of complexity. If users want non-unicode filenames, we could eventually support this.

If we encounter a non-unicode path, we could:

1. Return an error
2. Print a warning to the console
3. Display something like "N files with non-unicode paths omitted" in the directory listing

1 seems extreme, 2 seems reasonable but users might not look at the console and wonder where their non-unicode files are, and 3 is more obvious but would bother users who don't care and just want to see the files that they can get. I'm leaning towards 2. What do you think?

src/request_handler.rs
```diff
@@ -59,6 +69,16 @@ impl RequestHandler {
 
     Ok(Response::new(Body::from(body.into_string())))
   }
+
+  async fn serve_file(&self, file_path: &str) -> Result<Response<Body>> {
+    // todo: stream files
+    let file_contents = tokio::fs::read(dbg!(self
```
casey, line 75: dbg statement

(Actually it probably isn't necessary to mention stuff that will be caught by the forbid script)

src/request_handler.rs
casey: Agree that it's bad. I think we can open an issue and leave it for another PR. The person running the binary controls the contents of the www dir, so they can just not put large files in there for now :P

When we do want to do streaming, it looks like we'll want `Body::channel`, which returns a channel into which bytes can be written, or `Body::wrap_stream`, if we have a `Stream` implementation that will return chunks of the file.

src/request_handler.rs
```diff
@@ -92,63 +147,18 @@ pub(crate) mod tests {
...
-    test(|port, _dir| async move {
-      assert_eq!(
-        reqwest::get(format!("http://localhost:{}", port))
-          .await
-          .unwrap()
-          .status(),
-        200
-      )
-    });
+    test(|url, _dir| async move { assert_eq!(reqwest::get(url).await.unwrap().status(), 200) });
```
casey, line 150: I noticed some calls to `as_str` that were mostly not necessary, so I removed them.

src/request_handler.rs
```diff
@@ -59,6 +69,43 @@ impl RequestHandler {
...
+
+  async fn serve_file(&self, file_path: &FilePath) -> Result<Response<Body>> {
+    // todo: stream files
+    let file_contents =
+      tokio::fs::read(dbg!(self.working_directory.join("www").join(file_path))).await?;
+    Ok(Response::new(Body::from(file_contents)))
+  }
+}
+
+#[derive(Debug)]
```
casey, line 81: I noticed that a bad file path could escape the `www` directory, so I wrote `FilePath` before I noticed `disallow_access_outside_of_www`, so maybe you wanted to leave it for later.

This wrapper sanitizes file paths. It uses `Path::components` and complains if there are anything but "normal" path components.

src/request_handler.rs
```diff
@@ -166,27 +176,108 @@ pub(crate) mod tests {
...
+  #[test]
+  fn disallow_empty_path_component() {
+    test(|url, _dir| async move {
+      assert_eq!(
+        reqwest::get(format!("{}//bar.txt", url))
+          .await
+          .unwrap()
+          .status(),
+        // TODO: This should return `400`
+        StatusCode::INTERNAL_SERVER_ERROR
```
casey, line 279: We should switch to an Error enum with a `status` method that returns the HTTP status code to return.

src/request_handler.rs
casey: This could also just be a function, instead of a type:

```
fn file_path_from_uri(uri: &Uri) -> &str
```

src/request_handler.rs
soenkehahn: Sounds good.

src/request_handler.rs
soenkehahn: Exactly. :)

src/request_handler.rs
soenkehahn: I vote for 2. and for deferring this to a future PR. See also #25.

casey:
Returning 404s depended on the StatusCode stuff in this PR, so I added some commits fixing the windows build, and implementing #32.

PR #30, opened 2021-04-29
Fail CI if forbidden words are present

casey:
I used ripgrep because it respects `.gitignore` by default, which I
don't think is easy to do with grep. Since this check isn't OS-dependent,
I made it run on Linux only, to avoid having to write the command to
install ripgrep for multiple OSes.

I usually call this command `bin/lint`, and put other checks in here, but
`forbid` seemed more descriptive at the moment.


.github/workflows/build.yaml
```diff
@@ -76,3 +76,9 @@ jobs:
 
     - name: Check Formatting
       run: cargo fmt --all -- --check
+
+    - name: Check for Forbidden Words
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        sudo apt-get install ripgrep
```
soenkehahn, line 83: I think we should do `sudo apt-get update` before this.

PR #37, opened 2021-05-01
Run smoke test in justfile

casey:
I made it a separate target, so it's still possible to run the tests
with `just test`, even if something else is on port 8080.


soenkehahn (approved):
LGTM.

PR #40, opened 2021-05-01
Listen on loopback address during tests

casey:
This is a bit weird, since there's no good way to test that we're
actually listening on 0.0.0.0.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -26,7 +26,7 @@ impl Server {
         input: arguments.address.clone(),
       })?
       .next()
-      .ok_or_else(|| Error::AddressResolutionNoAddresses {
+      .ok_or(Error::AddressResolutionNoAddresses {
         input: arguments.address,
       })?;
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,11 +1,11 @@
 use crate::{
   environment::Environment,
-  error::{self, Result},
+  error::{self, Error, Result},
   request_handler::RequestHandler,
 };
 use hyper::server::conn::AddrIncoming;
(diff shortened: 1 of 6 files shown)


soenkehahn (approved):
I would've thought we would pass in e.g. `--address 127.0.0.1` during tests. (And implement `--address` first.) But I think this is also fine, since we do have a smoke test.

casey:
That sounds fine to me. I implemented `--address`. Looking up an address is somewhat complicated, because `ToSocketAddrs` does host resolution, and so might produce an I/O error. Also, the OS might return non-IP addresses, which are ignored, so the resulting iterator might be empty, which also needs to be handled.

I added a test that checks that the right error is produced with the right hostname.

src/server.rs
```diff
@@ -43,7 +53,7 @@ mod tests {
   use std::net::IpAddr;
 
   #[test]
-  fn listen_on_all_local_ip_addresses() {
+  fn listen_on_localhost() {
```
soenkehahn, line 56:
```suggestion
  fn listen_on_localhost_by_default_in_tests() {
```

tests/smoke.rs
```diff
@@ -7,7 +7,7 @@ use std::{
 };
 
 #[test]
-fn server_listens_on_localhost_8080() {
+fn server_listens_on_all_ip_addresses_port_8080() {
```
soenkehahn, line 10:
```suggestion
fn server_listens_on_all_ip_addresses_on_port_8080() {
```

soenkehahn (approved):

PR #41, opened 2021-05-02
Read files in 16KiB chunks

casey:
This creates a new type, `FileStream`, and implements `Stream` for it. Unlike `Future`, `Stream` requires manually implementing a trait. (I tried to find a crate, or implementation that I could copy, and surprisingly wasn't able to find one out there.)

I chose 16 KiB totally arbitrarily as the chunk size.

The chunk buffer is actually allocated on the stack:

```rust
let data = &mut [MaybeUninit::uninit(); 16 * 1024];
```

This should get optimized to just an increment of the stack pointer. It feels weird, but I don't actually think there are any negative effects of using 16KiB of stack space.

I wrote a unit test, but wasn't sure how to write an integration test. One way would be to restrict the amount of memory the process is allowed to use to, say, 16 MiB, using cgroups, and then stream a 32 MIB file. If it doesn't crash, it must be chunking it. Another form of test, since it's a performance optimization, would be to write a benchmark and then somehow make sure we don't regress on the benchmark too badly. I'm leaning towards the latter. What's your philosophy on testing performance optimizations that don't produce observable changes?

One thing which isn't great is that we're creating a new allocation for every read. The `Bytes` constructor here will allocate on the heap:

```rust
Poll::Ready(Some(Ok(Bytes::copy_from_slice(buf.filled()))))
```

I found [this crate](https://docs.rs/stream-body/0.1.1/stream_body/) that specifically solves this problem, but it's written for tokio 0.2, which is a few versions behind, so we'd have update it if we want to use it. It doesn't look suuuuper complicated, so the update should probably be doable.

One other possibility that I just thought of is to try to implement `hyper::body::HttpBody`, which might allow avoiding the allocation. I think that's probably worth trying, it looks like it's a simple trait, and not much more complex than `Stream`.

To Do:
- [ ] ~~Try implementing `HttpBody` instead of `Stream` to avoid allocation~~ (I looked into how to do this, and it's nontrivial. I think we should avoid bothering until we know for sure that the allocation is a problem. It could easily turn out that the cost of allocating is dwarfed by the cost of reading the data from disk.)
- [ ] Write end-to-end test
- [ ] Profile or do research to determine optimal chunk size


src/file_stream.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let want = &[0x15; 200];
+
+    std::fs::write(&path, want).unwrap();
+
+    let file = File::open(path).await.unwrap();
+
+    let mut stream = FileStream::new(file, FilePath::new("foo.txt"));
+
+    let mut have = Vec::new();
```
soenkehahn, line 78: `s/have/result`

src/file_stream.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[pin_project]
+pub(crate) struct FileStream {
+  #[pin]
+  file: File,
+  path: FilePath,
+}
+
+impl FileStream {
+  pub(crate) fn new(file: File, path: FilePath) -> Self {
```
soenkehahn, line 27: Maybe this shouldn open the `File` itself?

casey:
From the build, it looks like `libc::mkfifo` isn't defined on windows, so perhaps that `s!` macro is doing some magic.

I wrote some simple tests to try to figure out what was going on. The first just tried to use FileStream with a FIFO, which hung: 

https://github.com/soenkehahn/foo/pull/41/commits/bcb5e6d56cdc98a9663131945fda00d2834fa5a7#diff-19cc0f65267ab7859051d82a7d6ec5687e8d55eda1964b545d8e066b04dcd3a9R97

The second tried to just read and write a FIFO in two different threads (no async) in the same process, which hung: 

https://github.com/soenkehahn/foo/pull/41/commits/b292de09b58c846fbf1eb1aeb6a6ee85849283a2#diff-19cc0f65267ab7859051d82a7d6ec5687e8d55eda1964b545d8e066b04dcd3a9R126

I then switched it so that the writer was in another process, and it didn't hang:

https://github.com/soenkehahn/foo/pull/41/commits/868ca4e155de772b2d95509751b90c377326aed5#diff-19cc0f65267ab7859051d82a7d6ec5687e8d55eda1964b545d8e066b04dcd3a9L134

I'm not totally sure what's going on, since it seems to me that it should work with multiple threads.

casey:
I moved the test using threads into the playground, and it doesn't hang:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=30e9a61d1ea3e474e48d53d460c824dc

I am guessing that there is some OS-specific FIFO behavior that we're running into on macOS.

soenkehahn:
Yeah, that's really weird. Could it be that file handles are used to identify which futures/threads to wake up? When those futures/threads read or write to file handles? And since the file handle is the same both for reading and writing -- since it's the same named pipe -- that doesn't work properly? As in one of them doesn't get woken up and that's why it doesn't terminate. And on macos that's how it works for futures *and* threads, but on linux that's how it works only for futures and *not* for threads? Grasping at straws here.

Is this maybe also just non-deterministic?

soenkehahn:
I just added a commit (5f461cb) with future based tests for `FileStream` that passes reliably on my machine. It uses `tokio::fs::write` instead of `AsyncWriteExt`. Not sure why one works while the other doesn't.

PR #43, opened 2021-05-02
Add watch recipe to justfile

casey:


soenkehahn (approved):
I'll probably won't use this much, since I'm used to running `entr`. But LGTM.

PR #44, opened 2021-05-05
Run clippy on all targets and features

casey:
I think we weren't running clippy on code behind `#[cfg(test)]`, this makes
it run on all targets, test and non-test code, and any features that we eventually
add.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -76,7 +76,6 @@ jobs:
     - name: Clippy
       run: cargo clippy --all-targets --all-features
 
-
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -74,7 +74,8 @@ jobs:
         cargo test --test smoke
 
     - name: Clippy
-      run: cargo clippy --all
+      run: cargo clippy --all-targets --all-features
+
```
soenkehahn, line 78:
```suggestion
```

PR #48, opened 2021-05-06
Add links to view files in browser

soenkehahn:


src/request_handler.rs
```diff
@@ -226,13 +230,12 @@ pub(crate) mod tests {
   }
 
   #[test]
-  fn listed_files_are_download_links() {
+  fn listed_files_have_download_links() {
     test(|url, dir| async move {
       std::fs::write(dir.join("www").join("some-test-file.txt"), "").unwrap();
       let html = get_html(&url).await;
-      guard_unwrap!(let &[a] = css_select(&html, "a").as_slice());
+      guard_unwrap!(let &[a] = css_select(&html, "a[download]").as_slice());
```
soenkehahn, line 237: TIL about [css attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors).

src/request_handler.rs
```diff
@@ -241,14 +244,24 @@ pub(crate) mod tests {
...
       assert_eq!(file_contents, "contents");
     });
   }
 
+  #[test]
+  fn listed_files_can_be_played_in_browser() {
+    test(|url, dir| async move {
+      std::fs::write(dir.join("www").join("some-test-file.txt"), "").unwrap();
+      let html = get_html(&url).await;
+      guard_unwrap!(let &[a] = css_select(&html, "a:not([download])").as_slice());
```
soenkehahn, line 260: And I also learned about the [css not pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:not). So this will ignore all the download links.

casey:
I pushed a commit that uses mime_guess to guess the correct content type.

I was thinking that perhaps we should switch the presentation of the "view" and "download" links.

I think of "view" as being the default action, i.e. when you click a link, the browser will display its content, and "download" as being a non-default action. So it would make more sense to me to have the name of the file be the "view" link, and having the second link, with the text "download", be the download link.

I think this would be intuitive for users, since, for example, if you had an html file in the `www` directory, clicking on the main link would navigate to that document, which I think is what most people would expect.

What do you think?

soenkehahn:
Yeah, sounds good to me.

PR #60, opened 2021-05-11
Sort html file listing alphabetically

soenkehahn:


src/request_handler.rs
```diff
@@ -241,6 +252,23 @@ pub(crate) mod tests {
     });
   }
 
+  #[test]
+  fn listing_is_sorted_alphabetically() {
+    test(|url, dir| async move {
+      let www = dir.join("www");
+      std::fs::write(www.join("b"), "").unwrap();
+      std::fs::write(www.join("c"), "").unwrap();
+      std::fs::write(www.join("a"), "").unwrap();
```
soenkehahn, line 261: The order of read directory entries is both platform and filesystem dependent. (See [here](https://doc.rust-lang.org/std/fs/fn.read_dir.html#platform-specific-behavior). This is the documentation for `std::fs::read_dir`, but I think it applies to `tokio::fs::read_dir` as well.) This test (with files `b`, `c` & `a`, written in that order) seems to fail consistently on my machine when not sorting the output of `read_dir`. (We could try to come up with a more robust test with e.g. [`proptest`](https://crates.io/crates/proptest), but I think that's not necessary.)

src/request_handler.rs
```diff
@@ -67,8 +68,8 @@ impl RequestHandler {
         }
         body {
           ul {
-            @while let Some(entry) = read_dir.next_entry().await? {
-              @let file_name = entry.file_name().to_string_lossy().into_owned();
+            @for entry in files {
+              @let file_name = entry.to_string_lossy().into_owned();
```
casey, line 72: What about `file_name{,s}`?

```suggestion
            @for file_name in file_names {
              @let file_name = file_name.to_string_lossy();
```

I think we can avoid calling `into_owned` because we're not creating a temporary.

src/request_handler.rs
casey: I'm fine without a more robust test.

src/request_handler.rs
```diff
@@ -87,6 +88,16 @@ impl RequestHandler {
     Ok(Response::new(Body::from(body.into_string())))
   }
 
+  async fn lookup_directory(path: &Path) -> io::Result<Vec<OsString>> {
```
casey, line 91: I'm not sure how I feel about `lookup_directory`, it don't sound to me like it returns the contents of the directory.

```suggestion
  async fn entry_file_names(path: &Path) -> io::Result<Vec<OsString>> {
```

casey:
Minor suggestions, otherwise looks good.

src/request_handler.rs
```diff
@@ -56,7 +57,7 @@ impl RequestHandler {
   }
 
   async fn list_www(&self) -> io::Result<Response<Body>> {
-    let mut read_dir = tokio::fs::read_dir(&self.directory).await?;
+    let files = RequestHandler::lookup_directory(&self.directory).await?;
```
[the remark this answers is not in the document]
soenkehahn: Sounds good.

src/request_handler.rs
soenkehahn: Hmm, not sure about `entry_file_names`. Does the `entry` refer to the directory? But I also don't like `lookup_directory`. Maybe `read_dir`? Or `read_dir_sorted`?

Issue #71, opened 2021-05-26
Fix --version and --help


casey:
For both, they should display the appropriate text and then exit with `0` exit code. We'll want to check the error kind for HelpDisplayed and VersionDisplayed and handle them specially.

PR #73, opened 2021-05-27
Fix `cargo install` dependency resolution

casey:
Depends on #71.  Switches to new resolver and adds necessary hyper features.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,4 +1,4 @@
-all: build test smoke clippy fmt-check forbid
+all: build test smoke clippy fmt-check forbid check-install
 
 build:
   cargo build --all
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -19,5 +19,11 @@ fmt-check:
 forbid:
   ./bin/forbid
 
+check-install:
```
soenkehahn, line 22: Should we add this to the `all` target?

soenkehahn (commented):
Other than my comment, LGTM!

justfile
casey: Definitely

PR #81, opened 2021-05-29
Add contributing document

casey:
The idea is to make it clear that if changes are contributed, they are licensed according to LICENSE, i.e., CC0. Similar language is in the Apache license.


soenkehahn (approved):
LGTM.

PR #82, opened 2021-05-29
Add some styling

soenkehahn:


ATTRIBUTION.md
```diff
@@ -0,0 +1,5 @@
+# static/feather-sprite.svg
+
+Downloaded from https://unpkg.com/feather-icons/dist/feather-sprite.svg,
+following this guide: https://github.com/feathericons/feather#svg-sprite.
+See license here: https://github.com/feathericons/feather/blob/master/LICENSE
```
soenkehahn, line 5: [These icons](https://github.com/feathericons/feather) are licensed under an MIT license. I'm not sure whether we can pull them into this repo, since it's a different license. This is my attempt at doing it while giving proper attribution.

Alternatively, we could also not use these icons, of course. Although they're meant to be freely used, so hopefully we can figure this out.

soenkehahn:
This PR makes listings look like this:

image

It does pull in an icon set from [here](https://github.com/feathericons/feather) for the download icon.

ATTRIBUTION.md
casey: I think that's okay. The things that we author are licensed under the CC0, and things that we don't author are available under their respective licenses. The work as a whole is subject to the union of the restrictions of the licenses of the parts, but each separately licensed component of the work can be separated out and used/redistributed/modified according to the terms of that part's license.

This is already the case for our dependencies, all of which are currently MIT, or MIT / Apache 2.0. (You can use `cargo-lichking` to see licenses, `cargo install cargo-lichking && cargo lichking list`.)

Most commonly used licenses don't have any restrictions that we'd have trouble satisfying. (E.g. some require that you include the license, or that you attribute the author, which are all pretty easy.) The only licenses that we should watch out for are the GPL and the AGPL. If we have a GPL or AGPLed dependency, things get more complicated.

I'm usually pretty lax about attribution in projects, because nobody really cares or enforces it, but I think that adding an `ATTRIBUTION.md` as you've done is the scrupulously correct thing to do.

casey:
Looking much better already!

A few ideas:

- What do you think about reducing the size of each row? There's a lot of padding, which makes fewer items visible in a given amount of space. (A good way to get a listing with a lot of items is to do `cargo run -- --directory .` to just list the project directory. Then you can browse src and `.git` for dirs with lots of files.)

- The rows with a download icon are slightly larger than rows without a download icon. For visual consistency, they should probably be the same size.

- What do you think about making the download icon a bit smaller? More like the height of the text itself. This would also allow the rows to be smaller.

src/files.rs
```diff
@@ -136,6 +137,21 @@ impl Files {
...
+    html! {
+    svg
+      width=(size)
+      height=(size)
+      fill="none"
+      stroke="black"
+      stroke-width=(2)
+      stroke-linecap="round"
+      stroke-linejoin="round" {
+        use xlink:href="/static/feather-sprite.svg#download" {}
```
casey, line 150: This is really cool, I had no idea that you could somehow use `#…` fragment specifiers to select an icon inside of an SVG file.

src/files.rs
```diff
@@ -136,6 +137,21 @@ impl Files {
     Ok(Response::new(Body::from(body.into_string())))
   }
 
+  fn download_icon(size: u8) -> PreEscaped<String> {
+    html! {
+    svg
+      width=(size)
```
casey, line 143: Apparently you can also set these attributes with CSS:

```css
.icon {
  width: 24px;
  height: 24px;
  stroke: currentColor;
  stroke-width: 2;
  stroke-linecap: round;
  stroke-linejoin: round;
  fill: none;
}
```

This isn't that important now, but if we have more than one icon, this would allow us to avoid repeating attributes on each icon, but using, .e.g., `.icon` / `.icon-large`, etc.

README.md
```diff
@@ -36,3 +36,8 @@ You can run the tests locally with `cargo test`.
 Pull requests are tested on github actions, with the workflow defined in `.github/workflows/build.yaml`.
 You can run approximately the same tests locally with `just all`.
 (See [just](https://github.com/casey/just).)
+
+## License
+
+This software is licensed under [cc0-1.0](https://choosealicense.com/licenses/cc0-1.0/),
+unless stated otherwise in the `ATTRIBUTION.md` file.
```
casey, line 43: What do you think about:

```suggestion
Agora is licensed under [the CC0](https://choosealicense.com/licenses/cc0-1.0) with the exception of third-party components listed in [`ATTRIBUTION.md`](ATTRIBUTION.md).
```

To make it clear that Agora is always licensed under the CC0, and that it's only dependencies that have different licenses.

src/files.rs
```diff
@@ -136,6 +137,21 @@ impl Files {
     Ok(Response::new(Body::from(body.into_string())))
   }
 
+  fn download_icon(size: u8) -> PreEscaped<String> {
```
casey, line 140: Does Maud have an HTML type, that's more specific than `PreEscaped<String>`?

src/files.rs
casey: Another benefit is that the HTML will be smaller, because there won't be repeated attributes.

src/files.rs
soenkehahn: No, not really. This is just what the `html!` macro returns. It has this type alias, though: https://docs.rs/maud/0.22.2/maud/type.Markup.html. I'll switch to that.

src/files.rs
soenkehahn: Interesting. I'm switching to that.

I think there's also some advantages to trying to use Rust's abstraction mechanisms as much as possible. (And ignoring duplication in the html.) But yeah, it's a bit too early to see how this should actually be done. For one icon it certainly doesn't matter much.

README.md
soenkehahn: Very good!

src/files.rs
soenkehahn: Yeah, I agree.

I just found this deprecation notice: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href. So I think this should be just `href`.

ATTRIBUTION.md
soenkehahn: Yeah, I didn't even think about our rust dependencies. And in the the release binaries both rust dependencies and these icons are included. Sounds good then.

soenkehahn:
Now the rows have a smaller height, the download icon is smaller, and all the rows have the same height:
image
What do you think?

src/files.rs
```diff
@@ -116,17 +117,17 @@ impl Files {
...
                 @if file_type.is_file() {
-                  " - "
-                  a download href=(encoded) class=("download") {
-                    "download"
+                  a download href=(encoded) class="download" {
+                    (Files::download_icon())
                   }
                 }
               }
+              li class="separator" {}
```
casey, line 130: I just thought of something, I think this should be an `hr` element. There's no difference in terms of what it will look like, but `hr` is correct semantically. This is good, for example, for screen readers, which do better with semantic markup. I think this could just be `hr;`, and the CSS rule could just be `hr { … }`, i.e. it styles all `hr` elements.

casey:
Nice, looks good to me, aside from the `hr` thing.

The only thing I don't particularly like is how the download links are far away on the right, and that there's lots of whitespace, but I can't think of an obvious way to fix that, so I think we should merge this and worry about that later.

src/files.rs
soenkehahn: This is inside a `ul` element, where `hr`s are not allowed.

I *do* think that this separator is super hacky, so I wish we had a better solution. But I don't know one.

soenkehahn:
> The only thing I don't particularly like is how the download links are far away on the right, and that there's lots of whitespace, but I can't think of an obvious way to fix that, so I think we should merge this and worry about that later.

Agreed. Both on that the icons are too far away and that there's no obvious way to fix it. :|

PR #83, opened 2021-05-30
Build and publish binaries for tags

casey:
This builds and publishes an archive containing binaries for every tag pushed to github. All tags not starting with a `v` are marked as pre-releases. This is a hack so that only new version tags are marked as actual releases. I'm going to see if I can match a regex like `v[0-9]+\.[0-9]+\.[0-9]+` instead. (Update: There is no way to match strings with regexes T_T)

Also, this switches to a target-based build matrix instead of an OS-based build matrix. This doesn't have any effect right now, but will make it easier to build and test for different targets in the future.

I was thinking about also switching to musl for the release binaries, but figured we could do that in a follow-up PR.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,12 +24,19 @@ jobs:
 
     strategy:
       matrix:
-        os:
-        - macos-latest
-        - ubuntu-latest
-        - windows-latest
+        target:
+        - x86_64-apple-darwin
+        - x86_64-pc-windows-msvc
+        - x86_64-unknown-linux-gnu
+        include:
+        - target: x86_64-apple-darwin
+          os:     macos-latest
+        - target: x86_64-pc-windows-msvc
+          os:     windows-latest
+        - target: x86_64-unknown-linux-gnu
+          os:     ubuntu-latest
 
-    runs-on: ${{matrix.os}}
+    runs-on: ${{ matrix.os }}
 
     env:
       RUSTFLAGS: "--deny warnings"
@@ -56,6 +63,7 @@ jobs:
 
     - name: Install Rust Toolchain Components
       run: |
+        rustup target add ${{ matrix.target }}
         rustup component add clippy rustfmt
 
     - name: Info
@@ -90,3 +98,23 @@ jobs:
       run: |
         cargo install --path .
         agora --version
+
+    - name: Package
+      id: package
+      if: startsWith(github.ref, 'refs/tags/')
+      shell: bash
+      env:
+        TARGET: ${{ matrix.target }}
+        REF: ${{ github.ref }}
+        OS: ${{ matrix.os }}
+      run: ./bin/package
+
+    - name: Publish
+      uses: softprops/action-gh-release@v1
+      if: startsWith(github.ref, 'refs/tags/')
+      with:
+        draft: false
+        files: ${{ steps.package.outputs.archive }}
+        prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -109,9 +109,9 @@ jobs:
         OS: ${{ matrix.os }}
       run: ./bin/package
 
-    - name: Pre-release Check
-      id: prerelease
-      run: cargo run -p prerelease -- --reference ${{ github.ref }}
+    - name: Prerelease Check
+      id: is_prerelease
+      run: cargo run --package prerelease -- --reference ${{ github.ref }}
 
     - name: Publish
       uses: softprops/action-gh-release@v0.1.5
@@ -119,6 +119,6 @@ jobs:
       with:
         draft: false
         files: ${{ steps.package.outputs.archive }}
-        prerelease: ${{ steps.prerelease.outputs.value }}
+        prerelease: ${{ steps.is_prerelease.outputs.value }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,11 +30,11 @@ jobs:
         - x86_64-unknown-linux-gnu
         include:
         - target: x86_64-apple-darwin
-          os:     macos-latest
+          os: macos-latest
         - target: x86_64-pc-windows-msvc
-          os:     windows-latest
+          os: windows-latest
         - target: x86_64-unknown-linux-gnu
-          os:     ubuntu-latest
+          os: ubuntu-latest
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -63,8 +63,9 @@ jobs:
 
     - name: Install Rust Toolchain Components
       run: |
-        rustup target add ${{ matrix.target }}
         rustup component add clippy rustfmt
+        rustup target add ${{ matrix.target }}
+        rustup default `cat rust-toolchain`-${{ matrix.target }}
 
     - name: Info
       run: |
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -109,17 +109,9 @@ jobs:
         OS: ${{ matrix.os }}
       run: ./bin/package
 
-    - name: Prerelease Check
+    - name: Pre-release Check
       id: prerelease
-      if: startsWith(github.ref, 'refs/tags/')
-      run: |
-        REF=${{ github.ref }}
-        VERSION=${REF#"refs/tags/"}
-        if [[ $VERSION =~ ^v[[:digit:]]+[.][[:digit:]]+[.][[:digit:]]+$ ]]; then
-          echo "::set-output name=value::false"
-        else
-          echo "::set-output name=value::true"
-        fi
+      run: cargo run -p prerelease -- --reference ${{ github.ref }}
 
     - name: Publish
       uses: softprops/action-gh-release@v0.1.5
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -109,12 +109,24 @@ jobs:
         OS: ${{ matrix.os }}
       run: ./bin/package
 
+    - name: Prerelease Check
+      id: prerelease
+      if: startsWith(github.ref, 'refs/tags/')
+      run: |
+        REF=${{ github.ref }}
+        VERSION=${REF#"refs/tags/"}
+        if [[ $VERSION =~ ^v[[:digit:]]+[.][[:digit:]]+[.][[:digit:]]+$ ]]; then
+          echo "::set-output name=value::false"
+        else
+          echo "::set-output name=value::true"
+        fi
+
     - name: Publish
       uses: softprops/action-gh-release@v0.1.5
       if: startsWith(github.ref, 'refs/tags/')
       with:
         draft: false
         files: ${{ steps.package.outputs.archive }}
-        prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        prerelease: ${{ steps.prerelease.outputs.value }}
       env:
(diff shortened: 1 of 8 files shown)


casey:
This should be ready to go. Check out [this tag](https://github.com/soenkehahn/agora/releases/tag/package-test) to see what a release looks like. This also changes the names of the CI jobs, so we'll also need to change the required checks to the new names.

.github/workflows/build.yaml
```diff
@@ -90,3 +99,23 @@ jobs:
...
         agora --version
+
+    - name: Package
+      id: package
+      if: startsWith(github.ref, 'refs/tags/')
+      shell: bash
+      env:
+        TARGET: ${{ matrix.target }}
+        REF: ${{ github.ref }}
+        OS: ${{ matrix.os }}
```
soenkehahn, line 110: I think it'd be nicer to pass these arguments in through command line flags to `./bin/package`. (Command-line parsing in `bash` is pretty bad, so maybe we should only consider that if we re-write `./bin/package` in a different language.)

.github/workflows/build.yaml
```diff
@@ -90,3 +99,23 @@ jobs:
...
+      run: ./bin/package
+
+    - name: Publish
+      uses: softprops/action-gh-release@v0.1.5
+      if: startsWith(github.ref, 'refs/tags/')
+      with:
+        draft: false
+        files: ${{ steps.package.outputs.archive }}
+        prerelease: ${{ !startsWith(github.ref, 'refs/tags/v') }}
```
soenkehahn, line 119: Yeah, this is a bit hacky. Also, we don't want random tags to be pre-releases, do we?

Should we maybe come up with a more specific tag format for releases? For example we could say that only tags in the format of `release-$VERSION` are being released. (I don't really have an urgent need for prereleases right now, so I wouldn't bother thinking about that.)

bin/package
```diff
@@ -0,0 +1,50 @@
...
+case $OS in
+  ubuntu-latest | macos-latest)
+    ARCHIVE=$DIST/$BIN-$VERSION-$TARGET.tar.gz
+    tar czf $ARCHIVE *
+    echo "::set-output name=archive::$ARCHIVE"
+    ;;
+  windows-latest)
+    ARCHIVE=$DIST/$BIN-$VERSION-$TARGET.zip
+    7z a $ARCHIVE *
+    echo "::set-output name=archive::`pwd -W`/$BIN-$VERSION-$TARGET.zip"
```
soenkehahn, line 48: What does the `-W` do?

.github/workflows/build.yaml
```diff
@@ -57,6 +64,8 @@ jobs:
     - name: Install Rust Toolchain Components
       run: |
         rustup component add clippy rustfmt
+        rustup target add ${{ matrix.target }}
+        rustup default `cat rust-toolchain`-${{ matrix.target }}
```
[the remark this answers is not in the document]
casey: I personally like `…` a little better, since it's the first form I learned. (Although it can't nest, which is a downside.) In this case I like that the backticks are visually distinct from `${{…}}`.

bin/package
```diff
@@ -0,0 +1,50 @@
...
+    EXECUTABLE=target/$TARGET/release/$BIN.exe
+    ;;
+esac
+
+echo "Copying release files..."
+mkdir dist
+cp \
+  $EXECUTABLE \
+  Cargo.lock \
+  Cargo.toml \
```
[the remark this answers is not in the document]
casey: `Cargo.lock` allows you to see exactly which versions of which dependencies were used to build the binary. `Cargo.toml` lets you see which dependencies are direct, and also has other useful metadata, like the repository link, our emails, etc.

bin/package
```diff
@@ -0,0 +1,50 @@
...
+set -euxo pipefail
+
+VERSION=${REF#"refs/tags/"}
+SRC=`pwd`
+DIST=$SRC/dist
+BIN=agora
+
+echo "Packaging $BIN $VERSION for $TARGET..."
+
+test -f Cargo.lock || cargo generate-lockfile
```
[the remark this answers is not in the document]
casey: Good catch. This is from Just's `bin/package`, and I have no idea why it's there in the first place.

bin/package
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+VERSION=${REF#"refs/tags/"}
+SRC=`pwd`
```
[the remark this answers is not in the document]
casey: SGTM

.github/workflows/build.yaml
```diff
@@ -90,3 +99,23 @@ jobs:
       run: |
         cargo install --path .
         agora --version
+
+    - name: Package
+      id: package
+      if: startsWith(github.ref, 'refs/tags/')
+      shell: bash
```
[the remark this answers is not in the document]
casey: It's not needed, we have `defaults: shell: bash` above. (On Windows I think otherwise the shell defaults to powershell.)

.github/workflows/build.yaml
casey: That's what I was thinking, command line parsing in bash is pretty bad, and using named environment variables with `set -u` is less error prone.

I think that `bin/package` should be rewritten in Rust. I've made multiple errors making changes to it. What do you think about leaving it in bash for now, and switching to rust some time after stir/cradle has its first relatively stable release?

bin/package
casey: `-W` makes `pwd` output the windows version of the path, which is needed so the publish step can locate the archive.

.github/workflows/build.yaml
casey: The thinking is like this:

We want a way to be able to package and publish tags to GitHub Releases. (They're not real "releases" in the colloquial sense, that's just what GitHub calls "collections of files associated with tags".)

This is useful for testing that the package and publish workflow is correct, and for making binaries available with some patch, for us or others to try out without needing to build from source.

There are two options for marking a GitHub Release as "not a real release". You can mark it as a draft, or as a prerelease. Prerelease seemed more appropriate to me, because non-release releases will never be "undrafted". So they're not really pre-releases, that's just the best of the bad options for marking archives as not real releases.

I updated the script to be more selective. Now there's a `Prerelease Check` step that makes sure that the git ref actually matches `vN.N.N` and doesn't just start with `v`.

casey:
I re-wrote the logic that checks if a release should be marked as a prerelease in Rust. It's obviously much more verbose, but it has tests, does argument parsing, and probably took the same amount of time as it would have taken to write it in Bash, given that I didn't have to look up regex syntax, and didn't have to test it manually.

Here's a tag that it correctly marked as a pre-release:

https://github.com/soenkehahn/agora/releases/tag/package-test

And here's a tag that it correctly marked as a release:

https://github.com/soenkehahn/agora/releases/tag/v1.0.0

.github/workflows/build.yaml
soenkehahn: :) Yeah, I was thinking about using `stir` for this too... sounds good to me.

bin/package
```diff
@@ -0,0 +1,50 @@
...
+
+echo "Packaging $BIN $VERSION for $TARGET..."
+
+test -f Cargo.lock || cargo generate-lockfile
+
+echo "Building $BIN..."
+
+case $OS in
+  ubuntu-latest | macos-latest)
+    cargo rustc --bin $BIN --target $TARGET --release
```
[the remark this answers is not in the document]
casey: I think that `rustc` is not necessary here, but is necessary in the windows invocation, because we're passing extra arguments. I'll switch to `cargo build` here.

bin/package
```diff
@@ -0,0 +1,47 @@
...
+DIST=`pwd`/dist
+BIN=agora
+
+echo "Packaging $BIN $VERSION for $TARGET..."
+
+echo "Building $BIN..."
+
+case $OS in
+  ubuntu-latest | macos-latest)
+    cargo rustc --bin $BIN --target $TARGET --release
```
casey, line 15:
```suggestion
    cargo build --bin $BIN --target $TARGET --release
```

bin/package
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
casey: I don't think we'd need `rust-script`, we could just make it a normal cargo crate inside the current workspace.

casey:
Is this ready to go? I think I got all your comments, but I might have missed some. (Assuming you're okay with adding `bin/package` in bash now, and rewriting it in Rust once `stir` has a stable release.)

bin/prerelease/tests/integration.rs
```diff
@@ -0,0 +1,42 @@
+use executable_path::executable_path;
+use std::{process::Command, str};
+
+fn stdout(reference: &str) -> String {
```
soenkehahn, line 4: TIOLI:
```suggestion
fn test(reference: &str) -> String {
```

.github/workflows/build.yaml
```diff
@@ -90,3 +99,26 @@ jobs:
...
+      if: startsWith(github.ref, 'refs/tags/')
+      env:
+        TARGET: ${{ matrix.target }}
+        REF: ${{ github.ref }}
+        OS: ${{ matrix.os }}
+      run: ./bin/package
+
+    - name: Pre-release Check
+      id: prerelease
+      run: cargo run -p prerelease -- --reference ${{ github.ref }}
```
soenkehahn, line 114:
```suggestion
      run: cargo run --package prerelease -- --reference ${{ github.ref }}
```

.github/workflows/build.yaml
```diff
@@ -90,3 +99,26 @@ jobs:
...
+    - name: Package
+      id: package
+      if: startsWith(github.ref, 'refs/tags/')
+      env:
+        TARGET: ${{ matrix.target }}
+        REF: ${{ github.ref }}
+        OS: ${{ matrix.os }}
+      run: ./bin/package
+
+    - name: Pre-release Check
```
soenkehahn, line 112:
```suggestion
    - name: Prerelease Check
```

Since that's how we spell it elsewhere.

.github/workflows/build.yaml
```diff
@@ -90,3 +99,26 @@ jobs:
...
+      id: prerelease
+      run: cargo run -p prerelease -- --reference ${{ github.ref }}
+
+    - name: Publish
+      uses: softprops/action-gh-release@v0.1.5
+      if: startsWith(github.ref, 'refs/tags/')
+      with:
+        draft: false
+        files: ${{ steps.package.outputs.archive }}
+        prerelease: ${{ steps.prerelease.outputs.value }}
```
soenkehahn, line 122: Should `value` be called something like `is_prerelease`?

Cargo.toml
```diff
@@ -9,6 +9,9 @@ resolver = "2"
 repository = "https://github.com/soenkehahn/agora"
 homepage = "https://github.com/soenkehahn/agora"
 
+[workspace]
+members = [".", "bin/prerelease"]
```
soenkehahn, line 13: This is kind of nice. I think having a new package for a small script seems a bit overkill, but I don't know a better way. And this is less intrusive than I thought it'd be.

I mentioned `rust-script` before, which makes scripts more self-contained, but then I wouldn't know how to write tests for those scripts. So this LGTM.

bin/prerelease/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "prerelease"
+version = "0.0.0"
+authors = ["Casey Rodarmor <casey@rodarmor.com>"]
+edition = "2018"
+publish = false
```
soenkehahn, line 6: This is nice.

.github/workflows/build.yaml
soenkehahn: I see. Thanks for the explanation. I think the word 'draft' would make slightly more sense, IMO. But as far as I can tell, draft releases aren't even publicly visible, so that would not be that useful for providing binaries for testing purposes. So prereleases it is.

soenkehahn (approved):
Left some more comment. Apart from those LGTM.

.github/workflows/build.yaml
casey: Changed step to `is_prerelease`

PR #84, opened 2021-05-30
Make ./bin/forbid fail when ripgrep is not installed

soenkehahn:
`bash` is crazy.

### DIFF
--- a/bin/forbid
+++ b/bin/forbid
@@ -2,6 +2,7 @@
 
 set -euxo pipefail
 
+which rg
 ! rg \
   --glob !bin/forbid \
   --ignore-case \


casey:
Looks good.

Wow that's annoying. I have a few `bin/forbid` scripts I need to update. How did you find this?

casey:
I feel like bash scripts are so bad that we should switch to writing everything in Rust, even super trivial things like this. Plus it would give us an excuse to use `stir`.

casey (approved):

soenkehahn:
Yeah, I agree. `bash` is bad, and also, it's very nice to write scripts in the same language than the rest of the project. One downside is that `rust`'s tooling for writing self-contained scripts could be better.

PR #85, opened 2021-06-01
Improve snafu visibility attribute

casey:
This changed since I first use it. It now no longer has to be a string.


soenkehahn (approved):
LGTM.

PR #103, opened 2021-07-06
Implement header logger as example for tower layers

soenkehahn:
Just wanted to push this as an example for tower layers. cc: @casey

### DIFF
--- a/layer-playground/src/lib.rs
+++ b/layer-playground/src/lib.rs
@@ -1,0 +1,60 @@
+use futures::future::{self, FutureExt};
+use http::{
+  header::{HeaderName, HeaderValue},
+  Request, Response,
+};
+use std::fmt::Debug;
+use std::future::Future;
+use std::task;
+use tower::{Layer, Service};
+
+pub struct LogHeaders;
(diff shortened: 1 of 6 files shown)


layer-playground/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+use std::future::Future;
+use std::task;
+use tower::{Layer, Service};
+
+pub struct LogHeaders;
+
+impl<S> Layer<S> for LogHeaders {
+  type Service = WithLogHeaders<S>;
+
+  fn layer(&self, inner: S) -> Self::Service {
```
casey, line 16: What is the `Layer` trait for? I can see what it's being used for, but it's not clear to me why it's something that comes with `tower`. It seems you could just  write a `WithLogHeaders::new` method that would do the same thing, no trait required.

soenkehahn: Yeah, not sure why this is useful. It does give a bit of a guard-rail to implement middlewares, but yeah, I'm not sure if this has other advantages.

PR #114, opened 2021-07-28
Use cargo-limit in justfile

soenkehahn:


justfile
```diff
@@ -1,15 +1,15 @@
...
 test pattern='':
-  cargo test --all --all-features {{pattern}}
+  cargo ltest --all --all-features {{pattern}}
 
 smoke +args="":
-  cargo test --test smoke {{args}}
+  cargo ltest --test smoke {{args}}
 
 clippy:
   cargo clippy --all-targets --all-features
```
casey, line 15: I wonder if this is useful:

```suggestion
  cargo lclippy --all-targets --all-features
```

casey:
Looks good to me. There's also`lclippy`, which might actually be useful, if there are any cascading clippy errors.

justfile
soenkehahn: Good idea. I assumed that didn't exist.

PR #122, opened 2021-08-01
Redirect from /files to /files/

soenkehahn:
- Redirect from /files to /files/
- DRY up redirection tests


casey:
Looks good, I like the deduplication

PR #129, opened 2021-08-05
Rename lnd-client to agora-lnd-client and make it and agora publishable on crates.io

soenkehahn:


Cargo.toml
```diff
@@ -31,6 +30,10 @@ serde_yaml = "0.8.17"
 termcolor = "1.1.2"
 tonic = "0.4.3"
 
+[dependencies.agora-lnd-client]
+path = "agora-lnd-client"
+version = "=0.0.0"
```
casey, line 35: Can this just be `0.0.0` (which is implicitly `^0.0.0`) instead of `=0.0.0`?

casey:
Minor question about the version spec, but otherwise looks good to me.

PR #136, opened 2021-08-10
Improve readme

casey:
I added a bunch of what seems like useful information to the readme, much of it cribbed from a [comment I made on HN](https://news.ycombinator.com/item?id=28111417).

We could think about making an operator's manual type document, which we could build with [mdbook](https://github.com/rust-lang/mdBook), but I put it in the readme for now. I don't actually think having a super long readme is that much of a problem, but it might get unweidly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,10 @@ paid: false
 
 Currently, `agora` charges the low low price of 1,000 satoshis for all paid files.
 
+## Development Agora Instances
+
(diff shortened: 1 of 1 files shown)


casey:
A lot of this stuff is addressing questions that people on HN seemed to have, which wallets to use, how to get money in and out of bitcoin (which is why I mention River), and how inbound liquidity works.

soenkehahn:
LGTM.

PR #141, opened 2021-08-11
Add lightning invoice hyperlink

sgeisler:
This makes using browser plugins like [kWh](https://github.com/fiatjaf/kwh) easier. Ideally there'd also be a [WebLN](https://webln.dev/) integration if JS is activated, but I didn't know where to best integrate it. So this is a stop-gap solution to make it a bit more ergonomic.

I did not set up a full testing environment and I am new to the HTML macro templating system. Hopefully it does what I want (check ran through at least :man_shrugging:)


src/files.rs
```diff
@@ -274,7 +274,9 @@ impl Files {
               ":"
             }
             div class="payment-request" {
-              (invoice.payment_request)
+              a href={"lightning:" (invoice.payment_request)} {
```
casey, line 277: Maud uses `(…)` to delimit rust expressions to insert into the HTML. So I think this should be:

```suggestion
              a href=(format!("lightning: {}", invoice.payment_request)) {
```

I didn't test it though.

casey (commented):
Nice, thanks for this! Added a suggestion, which you can try just merging by hitting the merge button next to the suggestion, for the syntax.

I'm not familiar with WebLN, how would we use it?

soenkehahn:
Thanks for the PR!

> I did not set up a full testing environment and I am new to the HTML macro templating system. Hopefully it does what I want (check ran through at least man_shrugging)

Btw, we have some so-called `slow-tests` that are guarded behind a feature. They download and spin up their own `bitcoind` and `lnd` instances for the tests. So (ideally, fingers crossed) you don't have to set up any test environment yourself. You can run all the tests with `cargo run --all-features` or with `just test` (if you have [just](https://github.com/casey/just)). As you can see on the CI runs, your change also broke one of the slow tests.

I do think we probably would like some test coverage for this. Maybe we can add an assertion to [this test](https://github.com/agora-org/agora/blob/57b0a72c60821c203bd0207dfb98794111525b0b/src/request_handler.rs#L1032..L1041)?

I should also understand how these lightning urls work, I'll do some research.

casey:
We tweaked the layout a little bit. We were a little worried that if we linkified the entire payment request, a user might click it and nothing would happen because they don't have a `lightning:` URI scheme handler installed, so we made it a separate link with the text `Open in wallet`. That way, if they don't have a wallet installed, they probably won't be confused when they click `Open in wallet` and it doesn't do anything.

Here's a screenshot: 
Screen Shot 2021-08-17 at 2 28 53 PM

What do you think?

casey:
I'm going to go ahead and merge this, since it's definitely an improvement. If you have any comments on the changes, let us know!

PR #149, opened 2021-08-14
Fix publish script

casey:
The publish script doesn't work if `agora-lnd-client` needs to be published, since the main crate dry run publish will fail if the version of `agora-lnd-client` on crates.io.

This PR removes the dry runs. We could do something fancier, but maybe we should just do the simple thing now, and wait for someone else to write a good `cargo` command that publishes workspaces.


soenkehahn (approved):
LGTM.

PR #150, opened 2021-08-14
Add information about payments to the readme and the manifest description

soenkehahn:


README.md
```diff
@@ -2,6 +2,8 @@
 
 `agora` serves the contents of a local directory, providing file listings and downloads over HTTP.
 For example, you can point it at a directory full of PDFs, allowing users to browse and view the PDFs in their web browser.
+It allows to charge users for file downloads through [Lightning Network](https://en.wikipedia.org/wiki/Lightning_Network) payments.
+For that to work `agora` has to be configured to connect to a working [LND](https://github.com/lightningnetwork/lnd) node.
```
casey, line 6:
```suggestion
If `agora` is connected to an [LND](https://github.com/lightningnetwork/lnd) node, it can be configured to require [Lightning Network](https://en.wikipedia.org/wiki/Lightning_Network) payments for downloads.
```

Constructs of the form "VERB to VERB" should often be "VERB VERBing", so "It allows to charge…" should be "It allows charging…". Although not always, and I'm not sure what the rules are 😅 For example, "It's good to run." is correct, but "It's good running" is incorrect

casey (commented):
Good catch. I would reword the readme bit.

PR #151, opened 2021-08-15
Mention release RSS feed in readme

casey:
I didn't know that this feed was generated automatically until a packager of `just` mentioned it was how he kept up with new releases. I don't think it's common knowledge, so I figured it would be useful to document in the readme.


soenkehahn (approved):
LGTM.

(In another PR) we could also consider pointing out the possibility to watch, but only get notified for release events on github.

casey:
Good idea, I'll make a note on the board for that.

PR #159, opened 2021-08-18
Add crate, build, and chat badges

casey:
I created a Telegram group, because it's really valuable signal if users join, since I suspect actual users are more likely to join a group, as oppose to a twitter like or retweet, which might not be by a user. I made you an admin too.

Also added a crates.io and build badge to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
     <img src="https://github.com/agora-org/agora/workflows/Build/badge.svg" alt="build" />
   </a>
   <a href="https://t.me/agorific">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,11 @@
-# `agora`
+<h1 align="center">Agora</h1>
+<br />
+<div align="center">
```
soenkehahn, line 3: This `align="center"` doesn't seem to work. At least in the preview on github.

casey: Try the [full-file preview](https://github.com/agora-org/agora/blob/9f7027a7902ef3de234624b55add9d9c000f6f83/README.md), sometimes the snippet preview is messed up.

soenkehahn (approved):
LGTM.

PR #165, opened 2021-08-19
List public instances in readme

casey:
I don't have super strong feelings about where in the readme this should go, so let me know if you think it should be somewhere else.


soenkehahn (approved):
LGTM.

PR #176, opened 2021-08-30
Disable backtraces for tests

soenkehahn:
This PR disables backtraces when running tests. Before this PR backtraces get printed out for most test failures, but they're actually not useful, at least most of the time. And they make it much harder to see the actual error message.


src/error.rs
```diff
@@ -21,67 +23,89 @@ pub(crate) enum Error {
   AddressResolutionIo {
     input: String,
     source: io::Error,
+    #[cfg(not(test))]
```
soenkehahn, line 26: At first I thought we could just conditionally switch backtrace printing on and off. But for most tests, backtraces are not printed through `Error::print_backtrace`, but e.g. by calling `unwrap()` on an error that contains a backtrace, which will then be printed out through the `Debug` impl. This PR just removes the `backtrace` field when compiling tests.

src/error.rs
```diff
@@ -200,24 +241,26 @@ impl Error {
...
-                    MAIN_SEPARATOR, MAIN_SEPARATOR, MAIN_SEPARATOR
-                  )))
-              }
-              None => false,
-            },
-          );
-        }))
-        .print_trace(backtrace, write_color)
-        .ok();
+    if cfg!(not(test)) {
```
soenkehahn, line 244: The diffs looks big, but it's just this added `if` wrapped around the function body. Nothing else changed.

src/tests.rs
```diff
@@ -825,20 +825,6 @@ fn requesting_paid_file_with_no_lnd_returns_internal_error() {
...
-    let status = reqwest::get(context.files_url().join(".hidden").unwrap())
-      .await
-      .unwrap()
-      .status();
-    assert_eq!(status, StatusCode::NOT_FOUND);
-  });
-
-  assert_contains(&stderr, "agora::files::Files::check_path");
-}
-
```
soenkehahn, line 841: This tests backtrace printing. And that doesn't work as a unit test anymore, since backtraces are disabled. So I moved this test into `tests/smoke.rs`, which tests the `agora` executable compiled in non-test mode.

tests/smoke.rs
```diff
@@ -1,42 +1,87 @@
 use executable_path::executable_path;
```
soenkehahn, line 1: `smoke.rs` is probably not the best name anymore. Any better suggestions?

casey:
I agree it's a good idea to disable backtrace printing, but thought it was a little unfortunate to have to annotate all the backtrace fields, as well as not have backtraces during tests if we need them, so did a little bit of a digging for an alternative way to do this.

Snafu expects `backtrace` field types to implement [`GenerateBacktrace`](https://docs.rs/snafu/0.6.10/snafu/trait.GenerateBacktrace.html), so we could create our own backtrace type that implements this trait:

```rust
#[derive(Debug)]
pub(crate) struct Backtrace {
  inner: Option<snafu::Backtrace>,
}

impl GenerateBacktrace for BacktraceWrapper {
  fn generate() -> Self {
    Self {
      inner: if cfg!(test) {
        Some(snafu::Backtrace::generate())
      } else {
        None
      },
    }
  }

  fn as_backtrace(&self) -> std::option::Option<&Backtrace> {
    self.inner.as_ref()
  }
}
```

This would avoid the annotations, and let us easily re-enable backtraces if we need them while debugging a test.

Another option is to use `Option<Backtrace>` for the type of the field, which has an [implementation of GenerateBacktrace provided by snafu](https://docs.rs/snafu/0.6.10/snafu/trait.GenerateBacktrace.html#impl-GenerateBacktrace-for-Option%3CBacktrace%3E). It checks RUST_LIB_BACKTRACE and then RUST_BACKTRACE to see if it should generate backtraces, so this would allow conditionally turning backtraces on and off. One disadvantage of this is that most people will run agora without the environment variable set, and so won't see backtraces on crashes, and will need to know to set the environment variable to get them.

soenkehahn:
Yeah, I like that idea with the backtrace wrapper. I pushed a new commit, implementing that. Could you look at it again, please?

src/error.rs
soenkehahn: Reverted this, since this is not needed with the new `Backtrace` wrapper.

tests/smoke.rs
casey: Maybe just `tests/integration.rs`?

casey (commented):
Sweet, looks good to me.

tests/smoke.rs
soenkehahn: Sounds good.

PR #182, opened 2021-09-28
Use generic font families

casey:
I was thinking that maybe we should just use `monospace` everywhere, for a couple reasons:

- On my machine, the font in `code` blocks in rendered markdown and the font in our directory listings are different.
- I read an article (I wish I could find it now) that suggested that you should just use generic font families (`sans-serif`, `serif`, etc) because those are usually the most optimized fonts on the system, and it's hard to do better with a custom font stack.

However, `monospace` is weird, and there's an argument that we should use `font-family: monospace, monospace;`, because if you just use `font-family: monospace;`, it's smaller: https://stackoverflow.com/questions/38781089/font-family-monospace-monospace.

We could use `font-family: monospace, monospace;` everywhere, and pull in `normalize.css` which also makes all elements like `tt` and `code` use `monospace, monospace`. Which would also be fine with me.

What do you think?


soenkehahn:
Browser's are such a mess...

Yeah, we should probably use `normalize.css`. Note that the generated html from the `.index.md` files uses `<code>` tags.

PR #183, opened 2021-09-29
Deduplicate common imports in common.rs

casey:
As I've been threatening to do, I deduplicated our imports, and put everything into `common.rs` that I thought would be reasonably unambiguous.

### DIFF
--- a/src/arguments.rs
+++ b/src/arguments.rs
@@ -1,9 +1,6 @@
-use http::uri::Authority;
-use std::path::PathBuf;
-use structopt::{
-  clap::{AppSettings, ArgGroup},
-  StructOpt,
-};
+use crate::common::*;
+
+use structopt::clap::{AppSettings, ArgGroup};
 
 #[derive(Debug, StructOpt)]
 #[structopt(
--- a/src/arguments.rs
+++ b/src/arguments.rs
@@ -1,5 +1,4 @@
 use crate::common::*;
-
 use structopt::clap::{AppSettings, ArgGroup};
 
 #[derive(Debug, StructOpt)]
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,0 +1,64 @@
+pub(crate) use std::{
+  convert::Infallible,
+  env,
+  ffi::OsString,
+  fs::{self, FileType},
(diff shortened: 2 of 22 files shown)


src/common.rs
```diff
@@ -0,0 +1,64 @@
...
+  io::{self, Write},
+  mem::MaybeUninit,
+  net::{SocketAddr, ToSocketAddrs},
+  path::{Path, PathBuf},
+  pin::Pin,
+  str,
+  sync::Arc,
+  task::{Context, Poll},
+  time::Duration,
+};
```
soenkehahn, line 17: Could this section be put into the following `use` statement? (I'm not really familiar with the leading `::` in imports.)

src/common.rs
```diff
@@ -0,0 +1,64 @@
...
+    Body, Request, Response, StatusCode,
+  },
+  lexiclean::Lexiclean,
+  maud::Markup,
+  serde::Deserialize,
+  snafu::{IntoError, ResultExt},
+  structopt::StructOpt,
+  tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    task,
```
soenkehahn, line 39: I wonder whether using `task::spawn` would be a bit confusing, since there's also `std::task` and `std::thread::spawn`. Maybe `tokio::task::spawn` would be better.

src/arguments.rs
```diff
@@ -1,9 +1,6 @@
-use http::uri::Authority;
-use std::path::PathBuf;
-use structopt::{
-  clap::{AppSettings, ArgGroup},
-  StructOpt,
-};
+use crate::common::*;
+
```
soenkehahn, line 2: Can we get rid of this newline, please?

src/tests/slow_tests.rs
```diff
@@ -1,7 +1,7 @@
 use super::*;
+
```
soenkehahn, line 2: Can we not add this newline, please?

soenkehahn (approved):
Overall, this looks nice. I left a few comments. Also: Will we get warnings if we redundantly import something, that's already imported through `common`?

PR #186, opened 2021-09-30
Add a favicon

casey:
I made a really simple favicon, just a blue circle. If anyone asks we can tell them that the blue represents Greece, for agora, and that the circle represents the circular bitcoin economy that we're trying to foster. (I.e. an economy where people both earn and spend in bitcoin.)

This PR adds it to the static file tree, and also adds a special route at `/favicon.ico` to serve it, so that it automatically gets picked up by browsers without needing a `<link rel="icon"…>` in our HTML.

It's 32x32, which seems to be the current recommendation.


casey:
Added a test, `get` checks that the status code is 200, and I assert that the content type is "image/x-icon".

PTAL.

soenkehahn (approved):
LGTM.

PR #198, opened 2021-10-08
Use rustflags consistently on Windows

casey:
I took the opportunity to clean up `bin/package` a little bit as well.

I pushed this as a tag to trigger an actions run, to check that this actually works: https://github.com/agora-org/agora/actions/runs/1322120956


casey:
It turns out the keys needed to be in `.cargo/config.toml`, not `Cargo.toml`. Trying again T_T

PR #201, opened 2021-10-12
Run package step on master

casey:


casey:
@soenkehahn I think this is untestable, `github.ref` is `refs/pull/201/merge` for this PR. I think the best we can do is merge this and check manually that it works.

PR #218, opened 2021-10-17
Update logo

soenkehahn:
This logo is also just a blue circle, but thicker, which I think looks slightly better. In the spirit of incremental improvements. TIOLI.


casey:
Looks good to me. Can you add a recipe to the justfile to convert `logo.svg` to `static/favicon.ico`, so that if we change the SVG file in the future, we can update the .ico file?

PR #223, opened 2021-10-19
Recommend passing `--locked` to `cargo install`

Kixunil:
`cargo install` ignores `Cargo.lock` without `--locked` which could lead to breakage or supply chain attacks.
Using `--locked` fixes it, so it's useful to recommend.


Kixunil:
Yeah, if `Cargo.toml` differs then it's honored - this is intended. That's an important reason to commit `Cargo.lock` and test with it. Especially if you managed to review the dependencies you avoid supply chain attacks. (Even if you didn't review them deeply, it should still help decrease their impact.)

PR #225, opened 2021-10-20
🧿

casey:
The thicker logo gave me idea that our logo could be a [nazar](https://en.wikipedia.org/wiki/Nazar_(amulet)). A nazar is an eye-shaped amulet and decorative motif that is believed to ward against the evil eye.

I did a simple version of it in SVG, with the existing border, and used what seem to be common proportions and coloring.

Pros:
- It's very recognizable
- It's blue
- It has its own emoji
- It's very simple
- The connotation of protecting against the evil eye is nice, especially in the metaphorical sense of guarding against bad luck and surveillance, and protecting the marketplace
- It's not only Greek, but definitely still a popular symbol there. I've seen it most myself in Turkey.
- It can still stand in for an O in "Agora", like so: A G 🧿 R A
- Solid color border makes it easy to adapt to different shapes. For android we can use circle as-is, for iOS we can just use a blue squircle in place of the outermost blue circle. 

Cons:
- It's an existing symbol. I'm not sure what this means in terms of trademark. We might not be able to trademark it, but on the other hand, that might be a benefit, since if so then nobody can trademark it. We could certainly trademark "A G 🧿 R A" though.
- It might look "generic" to people who are already familiar with the symbol
- It might have some cultural connotation I'm not aware of (I think not though, trinkets with these things are sold by the armful, and it's shared by a whole bunch of cultures, so it's not at all thought of as a holy or culturally important symbol)

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Popular wallets include:
 
 - [Wallet of Satoshi](https://www.walletofsatoshi.com/), a hosted wallet for Android and iOS.
 - [Strike](https://strike.me/), a hosted wallet for Android, iOS, and Google Chrome.
-- [Muun](https://muun.com/), a self-custodial wallet for iOS and Android.
+- [Phoenix](https://phoenix.acinq.co/), a self-custodial wallet for iOS and Android.
 - [Breez](https://breez.technology/), a self-custodial wallet for iOS and Android.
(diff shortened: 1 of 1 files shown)


casey:
Pretty sweet SVG diff viewer on GitHub, btw:

Screen Shot 2021-10-19 at 8 43 53 PM

PR #239, opened 2021-10-21
Add icon for Apple devices

casey:
Serve an icon for Apple devices to use when they save an agora instance web page to their home screen.

- It's 180x180px, which is the current maximum suggested size
- It is served at `/apple-touch-icon.png`, which Apple devices request automatically, like `/favicon.ico`
- It is rendered with the background set to the color of the outermost circle of the eye, because Apple devices add their own mask

I've read inconsistent things about whether  Android devices will also use this icon, but I can't test that myself. If Android devices pick it up, then we should make it 196x196px, since that's the size recommended for Android devices.

I removed the `viewBox` from the SVG, and changed all the dimensions to be in terms of percentages. This makes the SVG scale to the document size when you open it in a browser or image viewer, and also lets us use `--zoom Y` if we want a YxYpx image.

Here's what it looks like on my device:

IMG_2608

### DIFF
--- a/static/index.css
+++ b/static/index.css
@@ -1,3 +1,8 @@
+html {
+  margin-left: 1rem;
+  margin-right: 1rem;
+}
+
 body {
   font-family: sans-serif;
   margin-left: auto;


soenkehahn:
My android device does not pick this up. So LGTM.

casey:
I added a test, and fixed the build for 1.48. The code changed slightly, so take another look, but it's pretty minor so I'll go ahead and merge it.

PR #251, opened 2021-10-31
Allow to offer dynamic script output for download

soenkehahn:
@casey: I was wrong about how processes get executed on linux: The kernel is capable of running scripts and binaries (e.g. ELF binaries), but in the case of scripts, they *have* to start with `#!`. So the kernel will *not* fall back on any shell, if the hashbang is missing. You can run scripts that don't start with `#!` from e.g. `bash`, and that will work by falling back on some shell, but that behavior is implemented in the shells you're using for that. (And apparently the exact behavior (e.g. which shell to use) depends on the parent shell being used.)

So this PR right now implements the following logic: When the given script starts with `#!` it is executed unmodified. If it doesn't a hashbang line is added (`#!/usr/bin/sh`). Judging from the tests that seems to be the behavior we want, but it's a bit magical.

The PR still needs to be cleaned up a lot, but I wanted to get your feedback early.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,8 +2,9 @@ use crate::common::*;
 use std::collections::BTreeMap;
 
 #[derive(PartialEq, Debug, Deserialize, Clone)]
+#[serde(tag = "type", deny_unknown_fields)]
 enum VirtualFile {
-  Script { source: String },
+  script { source: String },
 }
 
 #[derive(PartialEq, Debug, Default, Deserialize)]
@@ -29,16 +30,18 @@ impl Config {
     }
     path.read_dir().context(error::FilesystemIo { path })?;
     let mut config = Self::default();
-    for path in path.ancestors() {
+    let mut ancestors = path.ancestors().collect::<Vec<_>>();
+    ancestors.reverse();
+    for path in ancestors {
       if !path.starts_with(base_directory) {
-        break;
+        continue;
       }
       let file_path = path.join(".agora.yaml");
       match fs::read_to_string(&file_path) {
         Ok(yaml) => {
-          let parent =
+          let child: Config =
             serde_yaml::from_str(&yaml).context(error::ConfigDeserialize { path: file_path })?;
-          config.merge_parent(parent);
+          config.merge_child(child);
         }
         Err(error) if error.kind() == io::ErrorKind::NotFound => {}
         Err(source) => return Err(error::FilesystemIo { path: file_path }.into_error(source)),
@@ -47,11 +50,11 @@ impl Config {
     Ok(config)
   }
 
-  fn merge_parent(&mut self, parent: Self) {
+  fn merge_child(&mut self, child: Self) {
     *self = Self {
-      paid: self.paid.or(parent.paid),
-      base_price: self.base_price.or(parent.base_price),
-      files: self.files.clone(),
+      paid: child.paid.or(self.paid),
+      base_price: child.base_price.or(self.base_price),
+      files: child.files.clone(),
     };
   }
 }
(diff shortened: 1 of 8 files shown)


casey:
> @casey: I was wrong about how processes get executed on linux: The kernel is capable of running scripts and binaries (e.g. ELF binaries), but in the case of scripts, they _have_ to start with `#!`. So the kernel will _not_ fall back on any shell, if the hashbang is missing. You can run scripts that don't start with `#!` from e.g. `bash`, and that will work by falling back on some shell, but that behavior is implemented in the shells you're using for that. (And apparently the exact behavior (e.g. which shell to use) depends on the parent shell being used.)

I think I'm still a little confused about what the behavior is, and where the inconsistency is from. What you're saying is when you do `./foo` in `bash` and `foo` is a text file that doesn't start with a `#!`, then it's `bash` that chooses the shell to run it with, not the kernel? What happens if you try to run it with std::process::Command::new("./script")? (I.e. with no intermediate shell.)

On macOS, running a text file with no shebang with `std::process::Command::new` seems to work. I think it uses `/bin/sh`, but I'm not sure how it makes that decision. (`/bin/sh` on macOS is `bash`.)

I think adding a shebang automatically might not be desirable. It would make it unclear what was running a script without reading documentation, and it would mess up line numbers in error messages. Also if a user misspelled a shebang, i.e. added `!#/usr/bin/env`, we would add our own, which would be confusing.

I think it would probably be okay to just chmod +x the script and run it, and let the user add the `#!` if their system requires it.

src/config.rs
```diff
@@ -1,10 +1,19 @@
 use crate::common::*;
+use std::collections::BTreeMap;
+
+#[derive(PartialEq, Debug, Deserialize, Clone)]
+#[serde(tag = "type", deny_unknown_fields)]
+pub(crate) enum VirtualFile {
+  #[allow(non_camel_case_types)]
+  script { source: String },
```
casey, line 8: We probably want:

```suggestion
  #[serde(rename_all = "snake_case")]
  Script { source: String },
```

src/virtual_file.rs
```diff
@@ -0,0 +1,31 @@
...
+        source.to_string()
+      } else {
+        format!("#!/usr/bin/sh\n{}", source)
+      };
+      let tempdir = TempDir::new().expect("fixme");
+      let script_file = tempdir.path().join("script");
+      tokio::fs::write(&script_file, source).await.expect("fixme");
+      run!(%"chmod +x", &script_file);
+      run!(%"cat", &script_file);
+      let output = tokio::process::Command::new(script_file)
```
casey, line 20: Not necessary to do now, but we should remember to check the status code of the script. Not sure what to do if it fails.

soenkehahn:
> I think I'm still a little confused about what the behavior is, and where the inconsistency is from. What you're saying is when you do `./foo` in `bash` and `foo` is a text file that doesn't start with a `#!`, then it's `bash` that chooses the shell to run it with, not the kernel? What happens if you try to run it with std::process::Command::new("./script")? (I.e. with no intermediate shell.)

I get:
```
[src/main.rs:2] std::process::Command::new("./foo").output() = Err(
    Os {
        code: 8,
        kind: Uncategorized,
        message: "Exec format error",
    },
)
```

> On macOS, running a text file with no shebang with `std::process::Command::new` seems to work. I think it uses `/bin/sh`, but I'm not sure how it makes that decision. (`/bin/sh` on macOS is `bash`.)

Weird.

> I think adding a shebang automatically might not be desirable. It would make it unclear what was running a script without reading documentation, and it would mess up line numbers in error messages. Also if a user misspelled a shebang, i.e. added `!#/usr/bin/env`, we would add our own, which would be confusing.
> 
> I think it would probably be okay to just chmod +x the script and run it, and let the user add the `#!` if their system requires it.

Yeah, good points. I think that sounds right.

PR #252, opened 2021-11-03
Remove unused dependencies (and update some)

soenkehahn:


casey:
Sweet, this is great. LGTM.

Were we just not using these? Does this help with compile times? How did you find them?

soenkehahn:
> Were we just not using these?

Yeah, apparently not.

> Does this help with compile times?

Not that I could measure.

> How did you find them?

Comment all dependencies, compile and comment them in again manually. :( `cargo udeps` didn't work and just reported all dependencies as used.

casey:
> > Does this help with compile times?
> 
> Not that I could measure.

The negative is that it didn't make compile times faster, the positive is that at least cargo is smart enough not to let unused dependencies impact incremental compile times.

This probably helps our from-scratch compile times, since cargo does seem to compile all dependencies, including unused ones.

> Comment all dependencies, compile and comment them in again manually. :( `cargo udeps` didn't work and just reported all dependencies as used.

Weird, since that's its raison d'etre.

PR #257, opened 2021-11-14
Dockerize Agora

SachinMeier:
TODO: 
- GitHub workflow 
- Make sure dockerfile uses env vars properly. is there a better way to copy in files for the server and LND auth? 
- Optimize size: I used Ubuntu as final image to avoid the Alpine static linking issues. If these can be fixed Alpine or scratch would be better. 
- Optimize build time: In builder, we should build dependencies in a pseudo project so it's a different layer and doesn't need to rebuild every time src changes. 
-


Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# WORKDIR /usr/src
+
+# RUN USER=root cargo new agora
+
+# COPY Cargo.toml Cargo.lock /usr/src/agora/
+
+WORKDIR /usr/src/agora
+
+# Prepare for static linking
+# RUN apt-get update && \
```
SachinMeier, line 13: For static linking in case we switch to Alpine

Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+#     rustup target add x86_64-unknown-linux-musl
+
+# Download and compile Rust dependencies (and store as a separate Docker layer)
+
+# copy local dependencies
+# COPY ./agora-lnd-client ./agora-lnd-client
+# COPY ./lnd-test-context ./lnd-test-context
+# COPY ./bin ./bin
+# build dependencies
+# RUN cargo build --release
```
SachinMeier, line 25: This block of code is a failed attempt to build dependencies first. It resulted in the final binary being a default hello world rust program, not agora.

Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+RUN groupadd $APP_USER \
+    && useradd -g $APP_USER $APP_USER \
+    && mkdir -p ${APP}
+
+COPY --from=builder /usr/src/agora/target/release/agora ${APP}/bin/agora
+COPY static .
+# copy lnd auth files
+COPY ${LND_DIR} ./${APP}/.lnd
+# copy in files
+RUN echo ${AGORA_FILES}
```
SachinMeier, line 46: Debugging line to see why agora dir is written to files instead of my local files

.env.sample
```diff
@@ -0,0 +1,4 @@
+AGORA_FILES=files
+AGORA_PORT=8080
+LND_RPC_AUTHORITY=
+LND_DIR=.lnd
```
SachinMeier, line 4: Need to add LND enabled var or change entry point

SachinMeier: I've tried out a few solutions. I think the best option is to hardcode the content directory as `files`, which matches the current URL structure. And the LND `tls.cert` and `invoices.macaroon` will be in the `.lnd` folder.

.dockerignore
```diff
@@ -0,0 +1,4 @@
+/target/
```
casey, line 1: This duplicates a lot of what is in the `.gitignore`. Can we avoid having an extra file, and somehow reuse the `.gitignore`?

.env.sample
```diff
@@ -0,0 +1,2 @@
+AGORA_PORT=8080
```
casey, line 1: Can we delete this file in favor of using `-e` to pass arguments?

.gitignore
```diff
@@ -1,3 +1,6 @@
 /target/
 /tmp/
 /www/
+.lnd
```
casey, line 4: Where does the `.lnd` file come from?

Dockerfile
```diff
@@ -0,0 +1,56 @@
+# 1: Build the exe
+FROM rust:1.56 as builder
```
casey, line 2: Is it possible to use rust stable as the builder? That way we don't have to update it, and new releases of rust should be backwards compatible.

Dockerfile
```diff
@@ -0,0 +1,56 @@
+# 1: Build the exe
```
casey, line 1: I think docker files are pretty readable without comments, so I'd vote to delete them.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+FROM rust:1.56 as builder
+
+# WORKDIR /usr/src
```
casey, line 4: Commented out code should be removed.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# RUN USER=root cargo new agora
+
+# COPY Cargo.toml Cargo.lock /usr/src/agora/
+
+WORKDIR /usr/src/agora
```
casey, line 10: What happens if we don't set the `WORKDIR`?

Dockerfile
casey: We can remove this, and worry about static linking when/if we make the switch.

Dockerfile
casey: Can't we just build the `agora` binary, and it will build all dependencies automatically?

Dockerfile
casey: Also, what does `as builder` do?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# copy local dependencies
+# COPY ./agora-lnd-client ./agora-lnd-client
+# COPY ./lnd-test-context ./lnd-test-context
+# COPY ./bin ./bin
+# build dependencies
+# RUN cargo build --release
+
+# Build the exe using the actual source code
+COPY . ./
```
casey, line 28:
```suggestion
COPY ./ ./
```

casey: Just because it's what I've seen in other docker files.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+RUN cargo build --release
+
+# Copy the exe and extra files ("static") to an empty Docker image
+FROM ubuntu as final
```
casey, line 32: What does `as final` do?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+# copy local dependencies
+# COPY ./agora-lnd-client ./agora-lnd-client
+# COPY ./lnd-test-context ./lnd-test-context
+# COPY ./bin ./bin
+# build dependencies
+# RUN cargo build --release
+
+# Build the exe using the actual source code
+COPY . ./
+RUN cargo build --release
```
casey, line 29: Can we do `cargo install --path .`? If we do that, it will be installed to some system directory, and we then don't need to copy the binary to run it.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+FROM ubuntu as final
+
+ARG APP=/usr/src/app
+ENV APP_USER=agorauser
+
+RUN groupadd $APP_USER \
+    && useradd -g $APP_USER $APP_USER \
+    && mkdir -p ${APP}
+
+COPY --from=builder /usr/src/agora/target/release/agora ${APP}/bin/agora
```
casey, line 41: Can we avoid this if we do `cargo install` instead of `cargo build`?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# Build the exe using the actual source code
+COPY . ./
+RUN cargo build --release
+
+# Copy the exe and extra files ("static") to an empty Docker image
+FROM ubuntu as final
+
+ARG APP=/usr/src/app
+ENV APP_USER=agorauser
```
casey, line 35:
```suggestion
```

We can just hardcode the user, and I would call the agora user and group both `agora`.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+RUN groupadd $APP_USER \
+    && useradd -g $APP_USER $APP_USER \
+    && mkdir -p ${APP}
```
casey, line 39: I think it's enough to just run `useradd --system agora`:

```suggestion
RUN useradd --system agora
```

casey (commented):
I don't know docker well enough to be able to give an informed review, but I added a bunch of questions and comments.

I think we shouldn't worry about optimizing build time or using alpine, that can all be done later once we land a basic dockerfile.

casey:
My initial thought was that the docker file would also include LND, so people could get started really easily. Without LND, I'm not sure if there's a huge advantage over just downloading release binaries, or installing rustup and doing `cargo install agora`.

However, we do need to dockerize Agora for Umbrel, so maybe this is working even if it isn't a huge improvement over `rustup` + `cargo install agora` and launching it manually. @SachinMeier and @soenkehahn, thoughts?

entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+source .env
+# if lnd enabled, attempt to connect
+if [ "${LND_ENABLED}" = "1" ]
+then ./bin/agora --directory files --http-port $AGORA_PORT --lnd-rpc-authority $LND_RPC_AUTHORITY --lnd-rpc-cert-path .lnd/tls.cert --lnd-rpc-macaroon-path .lnd/invoices.macaroon
```
yzernik, line 6: You could probably simplify the entrypoint if you specify environment variables as fallbacks in arguments.rs:
https://docs.rs/structopt/0.3.25/structopt/#environment-variable-fallback

https://github.com/agora-org/agora/blob/9e8b31ae75107492a5f2494ee80b2ddc30abf9c9/src/arguments.rs#L47-L63

casey: Is that better? You would have the same conditional logic, but it would be in the `agora` binary instead of outside of it. It have the downside of making it less clear how the `agora` binary is configured, since if you don't see `--lnd-rpc-authority` on the command line, it might be pulling it from the environment.

Dockerfile
SachinMeier: yes, but building everything together means everything has to be rebuilt (including dependencies) every time we change anything in src. If we can build dependencies independently, then we rebuild times are much shorter. I'll remove this for now and make it a TODO

.dockerignore
SachinMeier: i think docker looks at dockerignore and git looks at gitignore. Never seen one use the other

Dockerfile
casey: Sounds good! Feel free to open issues to track things, I think that's better than including TODOs in the source code.

.gitignore
SachinMeier: I added it in case users (myself included) put a .lnd file in their agora dir. Somehow, the Docker image has to import the lnd files (macaroon & tls) on launch.

Dockerfile
SachinMeier: as builder is just a naming of the layer. And yes I will switch to stable

Dockerfile
SachinMeier: again its just a name. no functionality

.dockerignore
casey: What about making `.dockerignore` a symlink to `.gitignore`? That would keep them in sync.

Dockerfile
casey: Is there a benefit to naming layers? I'm a total docker noob, so don't understand what it does.

Dockerfile
SachinMeier: we do need to copy it, because `builder` and `final` are two different containers/systems. they wont share any sys directories.

entrypoint.sh
SachinMeier: I will simplify this entrypoint, but i don't think it needs a change to the src.

Dockerfile
SachinMeier: Rust doesnt have a stable tag. I will use slim

SachinMeier: Slim seems to fail to build agora.

SachinMeier:
A general problem with this approach is that we are trying to ship a docker image with the files and LND auth already set. As its built, Umbrel and docker users would need to manually copy these files *into* the docker container as its running. I'm unsure how to do this but our aim should be to build a docker image that then runs with LND auth (TLS, macaroons, and IP:port) *and* actual file content passed in as env vars. 

The problem with that is getting a running docker to copy more files into itself.

entrypoint.sh
yzernik: Yeah I guess it depends where you prefer to have that logic.

yzernik:
> 



> A general problem with this approach is that we are trying to ship a docker image with the files and LND auth already set. As its built, Umbrel and docker users would need to manually copy these files _into_ the docker container as its running. I'm unsure how to do this but our aim should be to build a docker image that then runs with LND auth (TLS, macaroons, and IP:port) _and_ actual file content passed in as env vars.
> 
> The problem with that is getting a running docker to copy more files into itself. @casey

Another option is you can create a docker-compose.yml file that includes bitcoind/LND/agora. Then you can keep the Dockerfile simple, and also have an easy way to deploy everything together.

entrypoint.sh
SachinMeier: I removed the `LND_ENABLED` var and used whether or not `LND_RPC_AUTHORITY` is set to determine lnd on/off

entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# if lnd enabled, attempt to connect
+if [[ -z "${LND_RPC_AUTHORITY}" ]]
+then ./bin/agora --directory files --http-port $AGORA_PORT --lnd-rpc-authority $LND_RPC_AUTHORITY --lnd-rpc-cert-path .lnd/tls.cert --lnd-rpc-macaroon-path .lnd/invoices.macaroon
```
yzernik, line 5: Related to the other comment. I don't think you should reference `.lnd` directly. Rather, use environment variables for `TLS_CERT_PATH` and `MACAROON_PATH`:

```
--lnd-rpc-cert-path $TLS_CERT_PATH --lnd-rpc-macaroon-path $MACAROON_PATH
```

alternatively, use an environment variable for `LND_DIR`:

```
--lnd-rpc-cert-path $LND_DIR/tls.cert --lnd-rpc-macaroon-path $LND_DIR/invoices.macaroon
```

README.md
```diff
@@ -56,6 +56,31 @@ Inside the checked out repository, running `cargo build --release` will build `a
 
 From within the repository, you can also run, e.g., `cargo install --locked --path . --root /usr/local`, which will copy `agora` to `/usr/local/bin/agora`.
 
+## Running with Docker
+
+The `agora` Docker image can be pulled from `TBD`. 
+
+### Building Agora Docker Image
+
+The Docker image can also be built directly from within the repository. The Dockerfile will copy the `/files` and `/.lnd`directory into the container and serve all files in the `/files` directory and search the `.lnd` directory for the `tls.cert` and `invoices.macaroon` files. The `.lnd` directory is only needed if you plan on connecting an LND instance.
```
yzernik, line 65: Again, you probably don't want to copy the `.lnd` directory into the docker image that you build. The idea of a docker image is to build it once, and make it re-usable. That won't be possible if you have a specific `.lnd` baked into the image.

entrypoint.sh
yzernik: And then you can mount the `.lnd` directory as a volume at run time:
```
docker run -p 8080:8080 -v ~/.lnd:/lnd -e LND_DIR=/lnd -e AGORA_PORT=8080 -e LND_RPC_AUTHORITY=10.0.0.20:10009 agora:latest
```

.dockerignore
SachinMeier: We don't want them to be the exact same file. While they're similar we want to dockerignore `.git` and other extra files.

casey:
Was experimenting with this on a Macintosh, and made a couple of small changes:

- Added a newline to the end of the file (vim, or maybe rustfmt, does this automatically)
- Capitalized the `AS`s, since they're Docker keywords

.dockerignore
casey: Gotcha, that makes sense, sounds good.

Dockerfile
casey: Ahhh, gotcha. My lack of docker knowledge is showing.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN cargo build --release
+
+FROM ubuntu AS final
+
+ARG APP=/usr/src/app
+ENV APP_USER=agora
+
+RUN useradd --system ${APP_USER}
+
+COPY --from=builder /usr/src/agora/target/release/agora ${APP}/bin/agora
```
casey, line 22: Can we put this in `/usr/local/bin/agora`, so that it can be called by just running `agora`?

```suggestion
COPY --from=builder /usr/src/agora/target/release/agora /usr/local/bin/agora
```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR /usr/src/agora
+
+COPY . ./
+RUN cargo build --release
+
+FROM ubuntu AS final
+
+ARG APP=/usr/src/app
+ENV APP_USER=agora
```
casey, line 18: I think this doesn't need to be configurable.

```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+RUN useradd --system ${APP_USER}
```
casey, line 20:
```suggestion
RUN useradd --system agora
```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+COPY --from=builder /usr/src/agora/target/release/agora ${APP}/bin/agora
+COPY static .
+COPY ./.lnd ${APP}/.lnd
+COPY ./files ${APP}/files
+
+RUN chown -R $APP_USER:$APP_USER ${APP}
```
casey, line 27: I think using long options is nice for non-interactive use, as a form of documentation:

```suggestion
RUN chown --recursive agora:agora ${APP}
```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR ${APP}
+COPY ./entrypoint.sh ./
+COPY .env .
+RUN chmod a+x entrypoint.sh
+USER $APP_USER
```
casey, line 33:
```suggestion
USER agora
```

README.md
```diff
@@ -56,6 +56,31 @@ Inside the checked out repository, running `cargo build --release` will build `a
...
+
+Building the image: 
+```bash
+docker build -t agora:latest .
```
casey, line 69: For documentation purposes:

```suggestion
docker build --tag agora:latest .
```

entrypoint.sh
casey, line 5: Since podman and docker can be used to map host ports to container ports, can agora used a fixed port inside the container?

Ideally port 80, since that's the standard web service port. If that requires configuration, because it's a privileged port, than port 8080 is good to.

If that works, then `AGORA_PORT` doesn't need to be configurable.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM ubuntu AS final
+
+ARG APP=/usr/src/app
+ENV APP_USER=agora
+
+RUN useradd --system ${APP_USER}
+
+COPY --from=builder /usr/src/agora/target/release/agora ${APP}/bin/agora
+COPY static .
+COPY ./.lnd ${APP}/.lnd
```
casey, line 24: This fails on my machine, since I don't have a `.lnd` directory.

casey: I was thinking about different ways of doing this. Could the user mount the certificate and macaroon individually to known paths inside the container? I.e. mount the two files individually, instead of mounting the directory.

Dockerfile
SachinMeier: it's not meant to be configurable, I just wanted to use a variable since its used multiple times in the file.

entrypoint.sh
SachinMeier: This makes sense

Dockerfile
SachinMeier: I think we should use a docker-compose file to do this at runtime, yes

yzernik: @SachinMeier You can also mount volumes at runtime with just the docker run command:

```
docker run -v ~/.lnd:/lnd -e LND_DIR=/lnd ...
```

casey: Is it possible to mount individual files? It might be nice to just mount the needed files, instead of a directory which must have a specific layout, and which might contain things that you don't want to pass through to the container.

yzernik: @casey Yeah I think that's possible:

```
docker run -v <HOST_MACHINE_FILE_PATH>:<CONTAINER_FILE_PATH> ...
```

neunenak: I would recommend against having a docker-compose file unless it's absolutely necessary, which it shouldn't be for agora's use case. docker-compose files make deployment more complex, and it's possible to mount individual files with just the `docker run` command.

casey:
Done in #291.

PR #261, opened 2021-11-23
Display file sizes

neunenak:
Display the file size of files in the listing.


casey:
@soenkehahn This looks good to me, so please take a look when you get the chance.

This displays file sizes on the right-hand side of listings. We used the `humantime` library, which isn't great, but seemed better than rolling our own byte formatting.

SachinMeier:
I'm not very familiar with Rust, but I ran this branch locally, servings several different file types, downloaded files, and didn't see any bugs. With some MP3 files, the sizes calculated by MacOS differed from the size displayed on agora.

neunenak:
> I'm not very familiar with Rust, but I ran this branch locally, servings several different file types, downloaded files, and didn't see any bugs. With some MP3 files, the sizes calculated by MacOS differed from the size displayed on agora.

This PR uses this rust std library method: https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len to calculate the file size. This is different than the space of blocks on the non-volatile storage taken up by the file, which is what desktop OSs might be reporting.

casey:
Merged in master, and moved tests to `src/integration.rs`, where such tests should now live.

I think `humantime` is a bit weird, and want to replace it, but I don't think that should hold up this diff, so merging.

PR #265, opened 2021-11-25
Replace humantime in favor of custom trait

casey:
I dislike that `humantime` is fallible, even though the only source of fallibility is from passing it a negative number, so doesn't apply. I thought about contributing a patch, but the project seems to be inactive. Also, I wanted to tweak how file sizes are displayed.

With this patch, file sizes are displayed without a decimal point if they're exact, and with a single digit after the decimal point, which might be zero, if they aren't.


casey:
Whoops, messed up the branch name.

PR #266, opened 2021-11-25
Replace humansize in favor of custom trait

casey:
I dislike that `humansize` is fallible, even though the only source of fallibility is from passing it a negative number, so doesn't apply. I thought about contributing a patch, but the project seems to be inactive. Also, I wanted to tweak how file sizes are displayed.

With this patch, file sizes are displayed without a decimal point if they're exact, and with a single digit after the decimal point, which might be zero, if they aren't.


src/display_size.rs
```diff
@@ -0,0 +1,72 @@
...
+      .map(|(i, suffix)| (2u64.pow(i as u32 * 10), suffix))
+      .filter(|(power, _)| value >= power - 1)
+      .last()
+      .unwrap();
+
+    if self.0 % power == 0 {
+      write!(f, "{} {}", self.0 / power, suffix)?;
+    } else {
+      write!(f, "{:.1} {}", self.0 as f64 / power as f64, suffix)?;
+    }
```
soenkehahn, line 33: I would've though it made more sense that
1. sizes in bytes (`B`) always get rendered without decimal points
2. sizes in units above bytes always get rendered with exactly one digit after the decimal point.

That seems more consistent to me. TIOLI.

soenkehahn (approved):
Apart from one comment LGTM.

src/display_size.rs
casey: I think you're probably right. I think it's nice that exact figures don't have a decimal point, but people probably won't understand that, and even if they do, it probably isn't worth the visual inconsistency between exact and inexact values.

PR #269, opened 2021-11-27
Statically link release binaries

### LINKED ISSUE #111: Consider statically linking release binaries

terror:
Fixes #111

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -18,32 +18,27 @@ jobs:
     strategy:
       matrix:
         target:
-        - aarch64-unknown-linux-gnu
-        - armv7-unknown-linux-gnueabihf
+        - aarch64-unknown-linux-musl
+        - armv7-unknown-linux-musleabihf
         - x86_64-apple-darwin
         - x86_64-pc-windows-msvc
-        - x86_64-unknown-linux-gnu
+        - x86_64-unknown-linux-musl
         include:
-        - target: aarch64-unknown-linux-gnu
+        - target: aarch64-unknown-linux-musl
           os: ubuntu-latest
           cross_compiling: true
-          target_rustflags: '--codegen linker=aarch64-linux-gnu-gcc'
-        - target: armv7-unknown-linux-gnueabihf
+        - target: armv7-unknown-linux-musleabihf
           os: ubuntu-latest
           cross_compiling: true
-          target_rustflags: '--codegen linker=arm-linux-gnueabihf-gcc'
         - target: x86_64-apple-darwin
           os: macos-latest
           cross_compiling: false
-          target_rustflags: ''
         - target: x86_64-pc-windows-msvc
           os: windows-latest
           cross_compiling: false
-          target_rustflags: '--codegen target-feature=+crt-static'
-        - target: x86_64-unknown-linux-gnu
+        - target: x86_64-unknown-linux-musl
           os: ubuntu-latest
           cross_compiling: false
-          target_rustflags: ''
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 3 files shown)


casey:
Pushed a tag to get the package workflow to run: https://github.com/agora-org/agora/actions/runs/1509778051

casey:
We are statically linking glibc, which apparently doesn't like being statically linked:

https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged

These are our glibc targets:

- aarch64-unknown-linux-gnu T1
- armv7-unknown-linux-gnueabihf T2
- x86_64-unknown-linux-gnu T1

These are our non-glibc targets:

- x86_64-apple-darwin T1
- x86_64-pc-windows-msvc T1

There are musl versions of the glibc targets we're using:

- aarch64-unknown-linux-musl T2
- armv7-unknown-linux-musleabihf T2
- x86_64-unknown-linux-musl T2

T1 and T2 indicate the tier of each target. See [Platform Support](https://doc.rust-lang.org/nightly/rustc/platform-support.html) for more info.

I think we should switch our glibc targets for musl targets, so that we aren't statically linking glibc. @soenkehahn What do you think?

Aside from just changing the targets, we'll need to install the correct linkers and pass the right linker flags.

soenkehahn:
That stackoverflow answer you linked to sounds very much like we want musl instead of glibc. It'd be nice to be able to reproduce that misbehavior described in there. I tried running `agora` as built from this PR, with static glibc inside an alpine docker container, but that worked fine. I'm not sure how to trigger it to `dlopen` anything. We could of course just switch to musl, but it'd be nice to actually be able to point to an observable misbehavior that that would fix.

casey:
I think observable misbehavior is hard to trigger. You would have a binary with a statically linked glibc, which would then use `dlopen` to load glibc again. So the conditions in which it would fail are a bit exotic, like maybe it's statically linked against version A of the glibc, and then is run on a system where dlopen causes version B to be loaded, and they're incompatible with each other.

I'll switch this PR to use the `-musl` targets.

casey:
Switched to `-musl` targets. These will probably fail to build, because they don't have the right linker, or we need to install musl, or something.

casey:
[Pushed a tag.](https://github.com/agora-org/agora/actions/runs/1514179229)

casey:
Maybe mess with locale settings? I think by default, conversions don't happen, and thus `iconv` stuff isn't triggered if you have the default local.

soenkehahn:
> I think observable misbehavior is hard to trigger. You would have a binary with a statically linked glibc, which would then use `dlopen` to load glibc again. So the conditions in which it would fail are a bit exotic, like maybe it's statically linked against version A of the glibc, and then is run on a system where dlopen causes version B to be loaded, and they're incompatible with each other.

I think it's easier to trigger than what you describe: Using `dlopen` to load glibc should fail on any system that doesn't have a dynamic glibc, e.g. alpine.

But yeah, we would still have to get glibc to actually execute `dlopen`...

I'm fine with just switching to musl.

soenkehahn:
Should we run our integration tests against a statically built binary? So that they're closer to production?

I don't think this should necessarily block this PR though. Otherwise LGTM.

soenkehahn (commented):
Apart from my comment about integration tests, LGTM.

casey:
This is failing on a [tag I pushed](https://github.com/agora-org/agora/runs/4348759750).

The armv7-unknown-linux-musleabihf job is failing because it's trying to call arm-linux-musleabihf-gcc, which isn't installed.

[This article](https://jensd.be/1126/linux/cross-compiling-for-arm-or-aarch64-on-debian-or-ubuntu) makes me think that Ubuntu doesn't include a toolchain for ARM + musl in apt.

We could download a cross-compiling toolchain here: https://musl.cc/

Or consider switching to rust-cross, which supports a bunch of targets at the expense of adding some complexity and not really being sure what's going on: https://github.com/rust-embedded/cross

casey:
This workflow which appears to build for arm musl targets might have some clues: https://github.com/nicolas-van/rust-cross-compile-example/blob/main/.github/workflows/rust.yml

terror:
PR is moved to #283

PR #273, opened 2021-11-29
Create a virtual file system abstraction

neunenak:
This commit is the first stab at creating a virtual filesystem abstraction. Eventually, `vfs.rs` is the only place in the codebase that should use APIs that directly access the file system. Other components of the system should go through the VFS to access file data and list the contents of directories, letting the VFS take care of reading the configuration file hierarchy, ignoring hidden files, and similar bookkeeping.

### DIFF
--- a/src/files.rs
+++ b/src/files.rs
@@ -27,50 +27,6 @@ impl Files {
     self.base_directory.join_file_path(path)
   }
 
-  fn check_path(&self, path: &InputPath) -> Result<()> {
-    if path
-      .as_ref()
-      .symlink_metadata()
-      .with_context(|| Error::filesystem_io(path))?
-      .file_type()
-      .is_symlink()
-    {
-      let link = fs::read_link(path.as_ref()).with_context(|| Error::filesystem_io(path))?;
-
-      let destination = path
-        .as_ref()
-        .parent()
-        .expect("Input paths are always absolute, and thus have parents or are `/`, and `/` cannot be a symlink.")
-        .join(link)
-        .lexiclean();
-
-      if !destination.starts_with(&self.base_directory) {
-        return Err(
-          error::SymlinkAccess {
-            path: path.display_path().to_owned(),
-          }
-          .build(),
-        );
-      }
-    }
-
-    if path
-      .as_ref()
-      .file_name()
-      .map(|file_name| file_name.to_string_lossy().starts_with('.'))
-      .unwrap_or(false)
-    {
-      return Err(
-        error::HiddenFileAccess {
-          path: path.as_ref().to_owned(),
-        }
-        .build(),
-      );
-    }
-
-    Ok(())
-  }
-
   pub(crate) async fn serve(
     &mut self,
     request: &Request<Body>,
@@ -108,7 +64,7 @@ impl Files {
       .with_context(|| Error::filesystem_io(path))?
     {
       let input_path = path.join_relative(Path::new(&entry.file_name()))?;
-      if self.check_path(&input_path).is_err() {
+      if self.vfs.check_path(&input_path).is_err() {
         continue;
       }
       let metadata = entry
--- a/src/files.rs
+++ b/src/files.rs
@@ -1,3 +1,4 @@
+use crate::vfs::Vfs;
 use crate::{common::*, file_stream::FileStream};
 use agora_lnd_client::lnrpc::invoice::InvoiceState;
 use maud::html;
@@ -6,6 +7,7 @@ use percent_encoding::{AsciiSet, NON_ALPHANUMERIC};
 #[derive(Clone, Debug)]
 pub(crate) struct Files {
   base_directory: InputPath,
+  vfs: Vfs,
   lnd_client: Option<agora_lnd_client::Client>,
 }
 
@@ -15,6 +17,7 @@ impl Files {
     lnd_client: Option<agora_lnd_client::Client>,
   ) -> Self {
     Self {
+      vfs: Vfs::new(base_directory.clone()),
       base_directory,
       lnd_client,
     }
@@ -183,7 +186,6 @@ impl Files {
   }
 
   async fn serve_dir(&self, tail: &[&str], dir: &InputPath) -> Result<Response<Body>> {
-    let config = self.config_for_dir(dir.as_ref())?;
     let body = html! {
       ul class="listing" {
         @for (file_name, file_type, bytes) in self.read_dir(dir).await? {
@@ -206,7 +208,8 @@ impl Files {
                 (bytes.display_size())
               }
             }
-            @if file_type.is_file() && !config.paid() {
+            // if vfs.paid(file)
+            @if file_type.is_file() && !self.vfs.paid(&dir.join_relative(file_name.as_ref()).unwrap()) {
               a download href=(encoded) {
                 (Files::icon("download"))
               }
--- a/src/files.rs
+++ b/src/files.rs
@@ -71,10 +71,6 @@ impl Files {
     Ok(())
   }
 
-  fn config_for_dir(&self, dir: &Path) -> Result<Config> {
-    Config::for_dir(self.base_directory.as_ref(), dir)
-  }
-
   pub(crate) async fn serve(
     &mut self,
     request: &Request<Body>,
@@ -240,14 +236,7 @@ impl Files {
     tail: &[&str],
     path: &InputPath,
   ) -> Result<Response<Body>> {
-    let config = self.config_for_dir(
-      path
-        .as_ref()
-        .parent()
-        .ok_or_else(|| Error::internal(format!("Failed to get parent of file: {:?}", path)))?,
-    )?;
-
-    if !config.paid() {
+    if !self.vfs.paid(path) {
       return Self::serve_file(path).await;
     }
 
@@ -259,7 +248,7 @@ impl Files {
     })?;
 
     let file_path = tail.join("");
-    let base_price = config.base_price.ok_or_else(|| {
+    let base_price = self.vfs.base_price(path).ok_or_else(|| {
       error::ConfigMissingBasePrice {
         path: path.display_path(),
       }
--- a/src/files.rs
+++ b/src/files.rs
@@ -6,7 +6,6 @@ use percent_encoding::{AsciiSet, NON_ALPHANUMERIC};
 
 #[derive(Clone, Debug)]
 pub(crate) struct Files {
-  base_directory: InputPath,
   vfs: Vfs,
   lnd_client: Option<agora_lnd_client::Client>,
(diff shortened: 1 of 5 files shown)


casey:
@soenkehahn We started working on a VFS abstraction. This just replaces a single call to `Config::paid` with a call to `self.vfs.paid`. The basic idea is:

- Initialize `Files::vfs` with `Vfs::new(base_directory)`
- Replace calls to `do_something_with(self.base_directory)` to `self.vfs.do_something()`
- Eventually remove `self.base_directory` and access everything through `self.vfs`
- Maybe move `vfs` into a crate, like `agora-vfs`, so that we can write and unit tests against the `vfs`, which would be much faster, since presumably the `vfs` crate wouldn't have a bunch of dependencies.

Currently, we have a config for each directory, which governs metadata about all the files in the directory. This seems like the wrong abstraction, and the `vfs` should let us query metadata for each file individually. Internally, the VFS might read and cache config files, if it's slow to read a config file for each file.

I thought it would probably be too much to do this at the same time as the dynamic content PR, both because that PR is already big, and because I'm pretty sure we want this, independently of the dynamic content PR.

soenkehahn:
Without having looked at the code, this sounds like a good plan to me!

neunenak:
I pushed a few more commits - now there is no code in `files.rs` that is directly touching the filesystem, and `check_path` can now be made private since no code outside `vfs.rs` uses it. This might be a good place to cut off a self-contained mergeable commit.

One thing I've noticed is that most of the methods on `Vfs` accept an `InputPath`. `file_type` is the only exception, and it could probably be fairly easily refactored to avoid this. I think we might want a model where `Files::serve` takes in `tail: &[&str]`, calls some method on `Vfs` to create an `InputPath` from it (which would do all error checking), and then works exclusively with `InputPath`, which is now guaranteed to point to a valid file. The `InputPath` can basically be an opaque handle to a file as far as the `Files` code is concerned.

casey:
> I pushed a few more commits - now there is no code in `files.rs` that is directly touching the filesystem, and `check_path` can now be made private since no code outside `vfs.rs` uses it. This might be a good place to cut off a self-contained mergeable commit.

This sounds like a great idea. If you think it's ready, go a head as mark it as ready for review. I'd start reviewing it now, but I don't want to get ahead of myself in case you have changes or things you'd like to clean up beforehand.
 
> One thing I've noticed is that most of the methods on `Vfs` accept an `InputPath`. `file_type` is the only exception, and it could probably be fairly easily refactored to avoid this. I think we might want a model where `Files::serve` takes in `tail: &[&str]`, calls some method on `Vfs` to create an `InputPath` from it (which would do all error checking), and then works exclusively with `InputPath`, which is now guaranteed to point to a valid file. The `InputPath` can basically be an opaque handle to a file as far as the `Files` code is concerned.

That sounds like a good idea. I think in a follow-up commit, we'll create something like `vfs::Path`, which is only for looking up files in the VFS. Once we do that, maybe we can get rid of `InputPath`, or only use it for filesystem paths, and not paths that come in from requests.

Just thinking out loud, this should definitely be a follow-up PR, but `vfs::Path` can do things like: Be UTF-8 only, reject anything that contains `.` or `..`, and always be absolute.

neunenak:
> That sounds like a good idea. I think in a follow-up commit, we'll create something like vfs::Path, which is only for looking up files in the VFS. Once we do that, maybe we can get rid of InputPath, or only use it for filesystem paths, and not paths that come in from requests.

Makes sense. I think if we have `vfs::Path` we don't need `InputPath`; or alternatively we could modify `InputPath` so that it does everything we want `vfs::Path` to do.

src/files.rs
```diff
@@ -201,20 +106,21 @@ impl Files {
...
             }
 
-            @if let Some(bytes) = bytes {
+            @if let Some(bytes) = entry.file_size {
               span class="filesize" {
                 (bytes.display_size())
               }
             }
-            @if file_type.is_file() && !config.paid() {
+            // if vfs.paid(file)
```
casey, line 114:
```suggestion
```

src/files.rs
```diff
@@ -201,20 +106,21 @@ impl Files {
...
+            @if entry.file_type.is_file() && !self.vfs.paid(&dir.join_relative(entry.file_name.as_ref()).unwrap()) {
```
casey, line 115: This makes me think that maybe the entry should say whether it's paid or not:

```suggestion
            @if entry.file_type.is_file() && !entry.paid {
```

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+  base_directory: InputPath,
+}
+
+impl Vfs {
+  pub(crate) fn new(base_directory: InputPath) -> Self {
+    Self { base_directory }
+  }
+
+  /// If an `.index.md` file exists in this directory, return its contents as a string.
+  pub(crate) fn directory_markdown_file(&self, dir_path: &InputPath) -> Result<Option<String>> {
```
casey, line 14:
```suggestion
  pub(crate) fn index_file_markdown(&self, dir_path: &InputPath) -> Result<Option<String>> {
```

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+      Ok(markdown) => Ok(Some(markdown)),
+      Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(None),
+      Err(source) => Err(Error::filesystem_io(&file).into_error(source)),
+    }
+  }
+
+  pub(crate) fn paid(&self, path: &InputPath) -> bool {
+    let config = Config::for_dir(
+      self.base_directory.as_ref(),
+      path.as_ref().parent().unwrap(),
```
casey, line 26: This unwrap should be removed. Paid can return `Result<bool>`.

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+    )
+    .unwrap();
+    config.paid()
+  }
+
+  pub(crate) fn base_price(&self, path: &InputPath) -> Option<Millisatoshi> {
```
casey, line 32: This unwrap should be removed, `base_price` can return `Result<Option<Millisatoshi>>`.

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+          path: path.as_ref().to_owned(),
+        }
+        .build(),
+      );
+    }
+
+    Ok(())
+  }
+
+  pub(crate) async fn read_dir(&self, path: &InputPath) -> Result<Vec<DirectoryEntry>> {
```
casey, line 105: I would call `DirectoryEntry` `DirEntry`, since that's what it's called in `std::fs`:

```suggestion
  pub(crate) async fn read_dir(&self, path: &InputPath) -> Result<Vec<DirEntry>> {
```

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+        .metadata()
+        .await
+        .with_context(|| Error::filesystem_io(&input_path))?;
+      let file_type = metadata.file_type();
+      let file_size = if metadata.is_dir() {
+        None
+      } else {
+        Some(metadata.len())
+      };
+      entries.push(DirectoryEntry {
```
casey, line 129:
```suggestion
      entries.push(DirEntry {
```

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+        file_size,
+        paid: self.paid(&path.join_relative(entry.file_name().as_ref()).unwrap()),
+      });
+    }
+    entries.sort_by(|a, b| a.file_name.cmp(&b.file_name));
+    Ok(entries)
+  }
+}
+
+pub(crate) struct DirectoryEntry {
```
casey, line 141:
```suggestion
pub(crate) struct DirEntry {
```

src/vfs.rs
```diff
@@ -0,0 +1,146 @@
...
+      let file_size = if metadata.is_dir() {
+        None
+      } else {
+        Some(metadata.len())
+      };
+      entries.push(DirectoryEntry {
+        file_name: entry.file_name(),
+        file_type,
+        file_size,
+        paid: self.paid(&path.join_relative(entry.file_name().as_ref()).unwrap()),
```
casey, line 133: this unwrap should be avoided

casey (changes requested):
I don't think that `Vfs` should make any assumptions about if a path that's being passed in is valid or not. I think it should assume that an `InputPath` might not exist, or was constructed outside of `Vfs`.

So I think that inside of `Vfs`, every function should start with a call to `check_path_recursive`, which the thing with `iter_prefixes`, and calls `check_path`.

Also, unwraps should be avoided, anything that unwraps can be made fallible.

neunenak:
I'm not entirely sure how notifications on this project are working, pinging you @casey since I updated the PR.

casey (approved):

casey:
Thanks for pinging me! I think I got the notification when you made the changes, but I was behind on responding. @soenkehahn if you have time to review, take a look. There's one thing that's kind of weird, which is that we only do the iter-prefix path check in one function, but that mirrors what we were doing before. I'm kind of inclined to merge it as-is, since all the tests pass.

casey:
LGTM, btw.

src/files.rs
```diff
@@ -203,20 +107,20 @@ impl Files {
               (file_name)
             }
 
-            @if let Some(bytes) = bytes {
+            @if let Some(bytes) = entry.file_size {
```
soenkehahn, line 110: TIOLI: rename `bytes` to `size` or `file_size`.

src/vfs.rs
```diff
@@ -0,0 +1,156 @@
...
+    self.check_path(&path)?;
+    let config = Config::for_dir(
+      self.base_directory.as_ref(),
+      path.as_ref().parent().ok_or_else(|| {
+        Error::internal(format!(
+          "Path {} has no parent",
+          path.display_path().display()
+        ))
+      })?,
+    )?;
```
soenkehahn, line 48: Can this by DRY'ed up with how the config is read in `paid`?

soenkehahn (approved):
> There's one thing that's kind of weird, which is that we only do the iter-prefix path check in one function, but that mirrors what we were doing before. I'm kind of inclined to merge it as-is, since all the tests pass.

I agree with this. It is a bit weird how checks are currently working, so there's room for further refactorings. But this is already much nicer than it used to be, so I'm in favor of merging.

I've left a few comments, apart from that LGTM. Very nice work!

casey:
The build failed with an error I saw on another PR, which was fixed by, I think, using a different branch with a different cache. I'm seeing if I can fix it by invalidating the cache.

casey:
I wanted to tinker with something that depends on this, namely a dedicated `vfs::Path` type, so I addressed the review comments and hit the merge button.

Issue #271, opened 2021-11-27
Normalize or reject URLs with `..` components


casey:
#270 made me think that we should normalize or reject URLs with paths containing `..`.

If user agents normalize or remove `..` from paths before making requests, then there doesn't seem to be a good reason for `agora` not to reject paths that are received with `..`.

casey:
Adding the ready to implement label, although that's just for researching what user agents do and finding the relevant spec. Suggestions for better names than "ready to implement" are most welcome.

casey:
My general feeling about `..` is that it is a nice affordance in user-facing interfaces, but a mistake anywhere else. Paths containing `..` should be normalized at the outermost layer, rejected if they can't be normalized away (i.e. leading `..` components), and not make their way further into the system.

casey:
Okay, I think that the [URI Generic Syntax RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) dictates that `.` and `..` should be normalized away by the user-agent before making the request, and shouldn't make it into a request.

PR #274, opened 2021-11-29
Update rust-embed version

### LINKED ISSUE #270: Path traversal in `/static` responder when running in debug mode
Steps to reproduce:

1. Run agora in *debug* mode, I used `cargo run -- --directory tmp --http-port 1234`
2. `curl --path-as-is https://localhost:1234/static/../../../../../../../../etc/passwd`
3. Get back the contents of your `/etc/passwd`

soenkehahn:
This fixes #270.


soenkehahn:
I would argue that we don't need a test for this, since we shouldn't test `rust-embed`'s behavior. Given that this is a bit scary, I could be convinced otherwise though. @casey

casey:
I agree. Also, I think if we do #271, i.e. normalize and reject `..` in the request handler, and write a test for that, we don't need to write a specific test for this vulnerability.

casey:
Also LGTM

PR #277, opened 2021-11-30
Refactor tests to use builder

casey:
This refactors the integration tests:

- Removes async code in favor of blocking code, since the tests don't need async, and blocking code is easier to write, and possibly compiles faster.
- Uses a builder pattern. `AgoraTestContext::builder()` creates a builder, which can then be customized with more method calls, and finally `builde()` constructs the test context. This should allow all customization to happen via the builder, which I find much easier to reason about vs the multiple nesting test functions we have currently
- Adds a bunch of helpers to `AgoraTestContext`, which allowed simplifying a bunch of tests quite a bit
- Moves some additional tests, although not many, into `tests/integration.rs`, which now use the builder pattern


casey:
@soenkehahn This is a gargantuan PR, so I wanted to get it up. I'll probably keep working on it, and start moving the slow tests and browser tests.

casey:
The bugfix_symlink_with_relative_base_directory is failing on windows. [Here's the log](https://github.com/agora-org/agora/runs/4426708988?check_suite_focus=true#step:8:174). I gave up on it for now because this is a massive PR already, and porting tests one by one from here on out seems like a better idea.

PR #278, opened 2021-12-05
Run CI on push

casey:


casey:
@soenkehahn I noticed that CI wasn't running on master after merges. This changes actions to run on PRs to all branches, and pushes to master.

PR #283, opened 2021-12-05
Statically link release binaries

terror:
This PR uncurses #269, we believe that there is something wrong with the cache.

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -35,14 +35,15 @@ jobs:
         - target: x86_64-apple-darwin
           os: macos-latest
           cross_compiling: false
-          target_rustflags: ''
+          target_rustflags: '--codegen target-feature=+crt-static'
         - target: x86_64-pc-windows-msvc
           os: windows-latest
           cross_compiling: false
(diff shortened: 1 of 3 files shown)


casey:
We gave up on on statically linking arm builds. x86 Ubuntu doesn't include an ARM + musl linker, so we would have had to use `cargo cross` or install the toolchain ourselves. cargo cross gave us weird errors, so we gave up on that. We could install a arm musl cross compilation toolchain (which would include a linker) ourselves, probably from https://musl.cc. But, we figured that we would much of the benefit of static linking just from doing it on Mac, Linux, and Windows, so that's what this PR does.

casey:
Pushed a [tag](https://github.com/agora-org/agora/actions/runs/1540571287).

casey:
Pushed another [tag](https://github.com/agora-org/agora/tree/cross-tag-4), which built, so LGTM.

@soenkehahn I'm inclined to merge this. It statically links more binaries, and arm is tricky, so not doing it for now seems like a good idea.

soenkehahn (approved):

PR #285, opened 2021-12-05
Update dependencies

soenkehahn:
Just ran `cargo update`. This gets rid of one `cargo audit` reported error (about `tokio`).


soenkehahn:
I bumped up the MSRV to `1.51` which is the first version that builds all updated dependencies.

casey:
Sweet, LGTM.

PR #286, opened 2021-12-06
Update proc-macro2 to 1.0.33, allow to run tests with cranelift

soenkehahn:
This PR allows to run the agora test-suite with cranelift. See https://github.com/bjorn3/rustc_codegen_cranelift/issues/1101.

One obvious advantage of cranelift is that compilation is faster, and one obvious downside is that executing the tests is slower. There may be other implications. So I'm not sure whether running the tests with cranelift is actually a good idea. In any case updating the `proc-macro2` version shouldn't hurt, and it allows us to experiment with cranelift.

See also https://github.com/bjorn3/rustc_codegen_cranelift#readme for a description on how to set up cranelift.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,7 +405,7 @@ version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "648ed8c8d2ce5409ccd57453d9d1b214b342a0d69376a6feda1fd6cae3299308"
 dependencies = [
- "proc-macro2 1.0.30",
+ "proc-macro2 1.0.33",
  "quote 1.0.10",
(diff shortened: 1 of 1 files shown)


casey:
Sweet! How did you do the selective version bump? I noticed that `Cargo.lock` changed without `Cargo.toml` changing, and I don't think I know how to do that for a single dependency.

soenkehahn:
> Sweet! How did you do the selective version bump? I noticed that `Cargo.lock` changed without `Cargo.toml` changing, and I don't think I know how to do that for a single dependency.

`cargo update -p proc-macro2:1.0.30`, where `1.0.30` is the *old* version, e.g. the version that you want to update.

PR #287, opened 2021-12-06
Move Config type into submodule of vfs

neunenak:
With the new VFS system, the `Config` type is now effectively an implementation detail of the VFS, and can exist as a non-public submodule of the `vfs` module.


neunenak:
@casey @soenkehahn this is something I noticed can be done now that the VFS code is merged. I'm not sure if there's a case for leaving `Config` as a type that should be available throughout the codebase.

casey:
This looks good to me. Is it possible to keep the vfs module in `vfs.rs` instead of `mod.rs`, or is it necessary to have submodules?

neunenak:
> This looks good to me. Is it possible to keep the vfs module in `vfs.rs` instead of `mod.rs`, or is it necessary to have submodules?

Just tried it and apparently that works, you can have `vfs.rs` and `vfs/config.rs` and the module hierarchy works out as expected.

casey:
Sweet, LGTM

For some reason the tests are super flaky, and I had to rerun them a bunch to get them to pass.

PR #295, opened 2022-03-03
Use 402 status code and update tests.

yzernik:
Use status code 402 when a file requires payment.

From [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402):

> The HTTP 402 Payment Required is a nonstandard response status code that is reserved for future use. This status code was created to enable digital cash or (micro) payment systems and would indicate that the requested content is not available until the client makes a payment.
> 
> Sometimes, this status code indicates that the request cannot be processed until the client makes a payment. However, no standard use convention exists and different entities use it in different contexts.

### DIFF
--- a/src/files.rs
+++ b/src/files.rs
@@ -221,7 +221,7 @@ impl Files {
       _ => {
         let qr_code_url = format!("/invoice/{}.svg", hex::encode(invoice.r_hash));
         let filename = invoice.memo;
-        Ok(html::wrap_body(
+        let mut response = html::wrap_body(
           &format!("Invoice for {}", filename),
           html! {
             div class="invoice" {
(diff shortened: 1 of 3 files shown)


casey:
We actually thought about this, but then elected not to do it. If I recall, our thinking was that there isn't any HTTP client software out there that will actually do anything useful in response to a 402. In particular, there isn't any software that will prompt the user to pay in response to a 402, and no way to deliver along with a 402 instructions for how to make the payment. In addition, some software, like HTTP clients, might misbehave if they get an unexpected status code. I'd definitely be open for returning 402s if it actually did something useful, but at the moment I don't think it's a benefit.

yzernik:
That makes sense.

In the future, it might be worth revisiting if the WebLN spec or browser extensions like Joule standardize 402 error handling.

PR #301, opened 2022-04-27
[readme] Remove defunct donation address

soenkehahn:
@casey: That address should be defunct, right?


casey:
It's an on-chain address, so we can still receive funds with it, and I'm guessing it's from our Lightning wallet, which I have the seed for, so if it received funds we could restore the wallet and get at the funds. I'm still fine with removing it though, since I haven't been checking it for funds and will probably forget to do so in the future.

casey (approved):

PR #304, opened 2022-05-08
Update dependency `lightning-invoice`

soenkehahn:
I'm not sure what's going on exactly, but I got build failures locally that are fixed by bumping the `lightning-invoice` version.


casey:
Weird, looks good to me.

yzernik:
@casey You probably have an older version of cargo/rustc installed.

The build only started breaking for me after I upgraded `rustc` from 1.58.1 to 1.60.0

